PR #25, opened 2020-01-26
Fix key issue on gallery item

konstilackner:


danielres:
good catch, thank you!! :+1:

PR #26, opened 2020-01-26
18 summary field

konstilackner:
- Add summary field on edit & create dream
- Update schema & resolver
- Display summary on DreamCard and on Dream Page

### DIFF
--- a/ui/components/DreamCard.js
+++ b/ui/components/DreamCard.js
@@ -58,7 +58,7 @@ export default ({ dream }) => {
         <div>
           <h3>{dream.title}</h3>
 
-          <p>{truncate(dream.description, 220)}</p>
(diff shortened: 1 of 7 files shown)


ui/components/DreamCard.js
```diff
@@ -58,7 +58,7 @@ export default ({ dream }) => {
         <div>
           <h3>{dream.title}</h3>
 
-          <p>{truncate(dream.description, 220)}</p>
+          <p>{dream.summary ? truncate(dream.summary, 200) : ''}</p>
```
danielres: the backend already limits to 200 characters, so this is not needed

danielres: however, I think it would be good to have a counter with the remaining available characters shown to the user in the UI

konstilackner: Yes, true. Will update.

PR #38, opened 2020-06-03
Fix/event membership dream

rootux:
# Fixes ✅
When a registered user tries to enter a dream of event he doesn't have membership in (Didnt join) - he get an error and can't enter that dream - This fixes it.

This fixes the following
Screen Shot 2020-06-03 at 12 32 02

# Alternative solution
An alternative (Perhaps better) fix would be to somehow set user.membership to default so when we check for fields such as `currentUser.membership.isAdmin` in the future we won't get an exception - Since I'm new into the code I'm not aware what's the best place to put it

# 🎩 Tophatting
Tested open an existing dream of a logged in use

### DIFF
--- a/.env.default
+++ b/.env.default
@@ -1,3 +1,5 @@
 MONGO_URL="mongodb://root:example@localhost:27017"
 JWT_SECRET="xxx"
 MAILGUN_API_KEY="xxx"
+PRETIX_URL="xxx"
+PRETIX_TOKEN="xxx"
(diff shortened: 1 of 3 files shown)


gustavlrsn:
Thanks for the PR! Really great to have you trying out the platform and contributing 🙂 

I think your solution here is a good one. You should be able to be logged in without having a membership for a particular event, and thus have the `currentUser.membership` field be `null` (the `.membership` field is queried based on the current event you are viewing).

PR #39, opened 2020-06-05
Extract away some of the domain coupling

rootux:
# Implements 
Extracted away some domain coupling to dreams domain and a specific email

# ⚠️ Warnings
⚠️ I did not had the ability to push into Vercel - therefore I was not able to test this.

I would suggest testing the following:
• Sending an email works as expected
• Getting an invite to join sends an email as expected

Testing should be done on **production**

⚠️ I can help and test this once I understand how to push into vercel


rootux:
Moving to main repo

PR #40, opened 2020-06-05
WIP: domain is now extracted into an env variable

rootux:
# Implements 
Extracted away some domain coupling to dreams domain and a specific email

# ⚠️ Warnings
⚠️ I did not had the ability to push into Vercel - therefore I was not able to test this.

I would suggest testing the following:
• Sending an email works as expected
• Getting an invite to join sends an email as expected

Testing should be done on **production**

⚠️ I can help and test this once I understand how to push into vercel


rootux:
@gustavlrsn Can you assist me here - Not sure I'm doing the `next.js` env right 😃

gustavlrsn:
@rootux removed the the `env` object in `now.json` to instead configure environment variables through Vercel's UI (also works through the `vercel env` command). This seems to be their new recommended approach, simplifies things a bit, and not using `vercel secrets` anymore.

To configure this you need to add these environment variables, either in the [UI or through the CLI command](https://vercel.com/docs/v2/build-step#environment-variables). Add for both **Production** and **Preview** (to make Preview deployments work too)
- `VERCEL_URL` with empty value, [a "system environment" variable that becomes the deploy url](https://vercel.com/docs/v2/build-step#system-environment-variables), and works dynamically with preview deployments etc.
- `MAILGUN_DOMAIN` (renamed this one just to make it more explicit since we're using two different domains for mailgun and deployment)
- `MAILGUN_API_KEY`
- `JWT_SECRET`
- `EMAIL_SENDER`

gustavlrsn:
@rootux updated the instructions above, since I realised that `VERCEL_URL` will always be a new `.now.sh` url per deployment, and not the actual domain that is pointed to it, i.e. not good for production (but it is good for making Preview deployments work with their own API endpoint)

PR #47, opened 2020-06-11
Added ability to set customFields on an object

rootux:
# Implements
* Added ability to set custom fields on an event
* Added JSON scalar to graphql
* Added ability to store custom field on a dream (And map them by object id)
* Added `storybook` to test out ui components separately (This should speed up ui development) 

# Top hatting 🎩
* Tested creating an event and setting custom fields
* Tested that only fields with values are rendered on guest view
* Tested that a user can modify custom fields

# Known issues
* After the event editor adds custom fields - A user needs to refresh the app for the custom events to appear - this is a general issue we have with data fetching method that needs to be addressed.

TODO
- [x] UI Styling - (Side note - I've used Grid but since then had a talk with @gustavlrsn where I understood that tailwind is the way moving forward) - @gustavlrsn 
- [ ] What about select value? Will be drafted and thought about
- [ ] Is required (For granting) - Needs to validate those custom field when user clicks the "apply for granting" button
- [x] Deleting a custom field does not update the UI and creates a broken view - @gustavlrsn can you look into that?
- [x]  UX - Deleting a custom field - Big warning with a YES / NO on the implications @rootux 
- [x] Bug - Checkbox - clearing it doesn't work @rootux 
- [x] Is Required bug for the admin - does not work @rootux 

# Future thoughts 
* Linked values - Make sure we can say - Only if that value is set to True - then show those values - For example - Only if a user select that he is going to use Electricity - we might want to show the required electricity questions
* Search should be done usign mongo atlas of elastic search (Currently custom fields are not searchable)

# Screenshots

Screen Shot 2020-06-25 at 16 40 26

Screen Shot 2020-06-25 at 16 41 24

Screen Shot 2020-06-13 at 21 54 27


rootux:
@gustavlrsn 
I've also added the Show on front page on setting the event.
It feels like we need some product work here.
Since this field would probably be limited to a few fields - a better way to represent this input is not on each of those but on a seperate place where you can simply add fields you want to be filtered by (With autocomplete) so
This is how it looks like ->
Screen Shot 2020-06-27 at 16 15 13

This is what I meant (Autocomplete) ->
Screen Shot 2020-06-27 at 16 17 09

rootux:
@gustavlrsn On the latest commit - now when I try and edit a certain field - it doesn't get updated on the ui and I need to refresh the page - I guess I need to get back the result and update the model in some way or am I missing anything?

To reproduce: Try to input into a dream a custom field. Then try to edit it - it gets updated but not on the ui

gustavlrsn:
I made the "event details" modal into a more general "Event Settings" modal with different tabs. We should move the Granting settings in here also (in another PR). 

As for the UI for adding and updating custom fields, I made it more "atomic" so that you only add or update one custom field at a time. I think this simplifies things a little bit in regards to not overwriting or editing things by mistake.
image

Also did a small UI tweak on when you fill out the custom field in a Dream, the text field or select field will take up the whole width instead.

gustavlrsn:
> @gustavlrsn On the latest commit - now when I try and edit a certain field - it doesn't get updated on the ui and I need to refresh the page - I guess I need to get back the result and update the model in some way or am I missing anything?
> 
> To reproduce: Try to input into a dream a custom field. Then try to edit it - it gets updated but not on the ui

I think it occured because the edit custom field mutation did not return the customFields, only the `id` of the Dream. As long as you return the id and whatever field you want to update, Apollo should take care of this through its normalized cache automatically.

gustavlrsn:
I'm seeing another bug I havn't looked in to yet. Can't change the value of a `Yes/No` question to `No`.

rootux:
> I made the "event details" modal into a more general "Event Settings" modal with different tabs. We should move the Granting settings in here also (in another PR).
> 
> As for the UI for adding and updating custom fields, I made it more "atomic" so that you only add or update one custom field at a time. I think this simplifies things a little bit in regards to not overwriting or editing things by mistake.
> image
> 
> Also did a small UI tweak on when you fill out the custom field in a Dream, the text field or select field will take up the whole width instead.

Looks super slick. Great work.

rootux:
@gustavlrsn Fixed the issue with the rendering and another issue of scrolling.
Also removed the Show on Front page (There is another pr that solves it)
Looks like we can merge this ✅

gustavlrsn:
> @gustavlrsn Fixed the issue with the rendering and another issue of scrolling.
> Also removed the Show on Front page (There is another pr that solves it.

Great!

> Only missing part is understanding what to do with the `Is required` field in terms of where does this comes into play - We can merge it without implementing it at the moment.
> Looks like we can merge this ✅

Sounds good, I'll merge this. Great job.

PR #50, opened 2020-06-16
Delete dreams

gustavlrsn:


gustavlrsn:
Changed this to do a hard delete of dreams (as opposed to the branch name), since I added validation so that you can only delete dreams that has not received any grants, to prevent having to check if a dream is deleted in a lot of other places. Retrieveing a soft deleted dream would require db access right now anyway. Can change this in the future if deemed necessary to do soft deletes instead.

Issue #54, opened 2020-06-29
Show on front page


rootux:
# Feature description
* Event owner can have an autocomplete box where he can set custom fields that will be shown on frontpage (See image 3)
* General User / Guest can toggle it using a label on the Front page - showing that specific custom field is set on a dream. So for example - clicking on "Help needed" label would filter data for all dreams, showing "Help needed".
* Start with only rendering a text field and a checkbox (as Yes/No)

# Questions

# Mock
As a guest / normal user - on front page of an event we would have those new `labels` -> 
<img width="508" alt="Screen Shot 2020-06-29 at 16 35 28" src="https://user-images.githubusercontent.com/1032526/86017793-25e0b700-ba2d-11ea-8ad0-46793cdf2a1e.png">

Filtered view of only dreams with the "Help needed" set - showing the help needed section
<img width="913" alt="Screen Shot 2020-06-29 at 17 52 06" src="https://user-images.githubusercontent.com/1032526/86020925-34c96880-ba31-11ea-8abe-e779d35e0c8d.png">

Image 3 - As an event owner -> On the custom fields we would have another autocomplete that would allow user to select custom fields. We would use material react multi select (Which we already have as an imported library)
<img width="765" alt="Screen Shot 2020-06-29 at 17 57 37" src="https://user-images.githubusercontent.com/1032526/86021640-fbddc380-ba31-11ea-8804-22b02fb9d1c4.png">

PR #55, opened 2020-06-30
Feat - Show on frontpage (aka filter labels)

rootux:
# Implements
* Solves issue #54 
* Added ability to filter by custom fields on the front page

# Tophatting 🎩
* Tested setting some fields on.
* Tested setting some fields off.
* Tested selecting those fields
* Added a custom field. Set it as filtered labels. Then removed it.

# Known issues
* Updating the filter labels (aka show on front page) require a refresh to be seen
* Design issue - Currently we are showing all dreams and only showing information where that field is set (Perhaps a better way is to hide some dreams)

# Screenshots
Screen Shot 2020-07-01 at 21 30 24

Screen Shot 2020-06-30 at 20 40 32
Screen Shot 2020-06-30 at 20 05 01
Screen Shot 2020-06-30 at 20 04 56


rootux:
@gustavlrsn Fixed the issue we had in our morning meeting

gustavlrsn:
I was trying out some ideas for design improvements and landed somewhat in this idea of just allowing you to choose any custom field you want to show on the card (and Summary being the default). 
image

We only have to show this dropdown/select input if there are any custom fields defined. This would require much less code to maintain, but make some of your work redundant. 

What do you think @rootux? It would make some of your work redundant, but also less code to maintain :-)

One potential downside could be if you for some reason do not want your users to be able to show a certain field on that card, but can't think of any reason why you'd necessarily care about that

rootux:
> I was trying out some ideas for design improvements and landed somewhat in this idea of just allowing you to choose any custom field you want to show on the card (and Summary being the default).
> 
> image
> 
> We only have to show this dropdown/select input if there are any custom fields defined.
> 
> What do you think @rootux? It would make some of your work redundant, but also less code to maintain :-)
> 
> One potential downside could be if you for some reason do not want your users to be able to show a certain field on that card, but can't think of any reason why you'd necessarily care about that

Yep. That hurt as it took a lof of time into that 😅
I'm totally fine with altering it to a different design (No attachment 😉 )
I'm not sure what feels better.
Perhaps we can check with @aerugo and @skoteskote what they think about it, design wise.

In terms of time waste - I would say let's wait for some more ux / user input to put more into that

rootux:
Add a label icon next to each filter label.
Added on top right a filter label icon

rootux:
Going with the older solution of choosing whether or not to show it

rootux:
New screenshots
Screen Shot 2020-07-13 at 14 45 20
Screen Shot 2020-07-13 at 14 45 17

Screen Shot 2020-07-13 at 13 30 03

gustavlrsn:
> New screenshots of the latest ui changes - @gustavlrsn
> 
> * Removed the auto complete box and went with your design of "Show on front page"

Great.

> * Added a label + on the dream card showing that label

The label is covering up the "unpublished" label if it exists. Also thinking that the tag/label icon might be confusing with the regular concept of tags? (which we probably will introduce at a later point)

Tried adding the label closer to the text, to indicate that you are viewing something other than the regular summary, what do you think?
image

This design also implies that we should filter away dreams that do not have that particular field set, and that would also bring these filter labels more into alignment with the rest of the controls (search, favorites) that also filter stuff.

rootux:
I believe it's done now - Filtering existing dreams on the client side
can be done better instead of O(N) * 2 - filtered once O(N)

Screen Shot 2020-07-24 at 18 43 48
Screen Shot 2020-07-24 at 18 43 52
Screen Shot 2020-07-24 at 18 44 52

@gustavlrsn Please merge

PR #60, opened 2020-07-07
Include subdomain in headers

gustavlrsn:


rootux:
That looks great. How do I use it?
Does that work for development also?

rootux:
Why not instead in `api/index.js` do the following:
const { getHostInfo } = require('./utils/getHostInfo');

then later on
`const { subdomain } = getHostInfo(req);`

gustavlrsn:
> Why not instead in do the following:
> 
> * Move `getHostInfo` module to the server side
> * In `api/index.js` add -
>   `const { getHostInfo } = require('./utils/getHostInfo');`
> * Then later on in that file
>   `const { subdomain } = getHostInfo(req);`

Yeah, that sounds cleaner, but I don't think it will work out of the box since the frontend currently talks to `http://localhost:3000/api` without the subdomain, so I don't think the information will be there in the request on the API side

rootux:
I see. Great. Merging.

Issue #58, opened 2020-07-06
Support for Organizations


rootux:
# Description
Implement Organizations as owners of `Events` and connected to `Users`
At the moment - Membership are on the event level. Users -> Events

# Tasks
- [x] Database changes (See next section below)
- [x] Set up ability to set domains[?] (Main domain - dreams.wtf Subdomains - borderland.dreams.wtf Custom domains)
- [x] When loading events - make sure we are loading events only for the specific organization
- [x] Handling setting `organization` through the url (To support guest users) - This might be an issue as borderland.dream.wtf can be aliased by someone to be borderland.com
- [x] Security - right now we are validating that a user `isOrgAdmin` - Now we should check which organization that user is assigned to and also if he is org admin.
- [x] Creating a user changes depending on which organization the user is currently at
- [ ] Create a backend user interface to allow Deleting an organization
- [ ] Create a backend user interface to allow Adding a new organization
- [ ] [Migrate existing users to a mother organization?] @aerugo @skoteskote ?

# Database changes
- [x] Change users model to holds organizationId which would link a user to
an organization. organizationId Can't be null.
- [x] Change event model to holds organizationId

> Organization
id - Mongo's Object id
name - String - Borderland
subdomain - String - For example borderland (As to borderland.dreams.wtf)

# Other notes

No need to see all organizations bird eye view
Linked to https://miro.com/app/board/o9J_ktbysWs=/

# Questions

~~1. Do we already have some admin backend interface @gustavlrsn ?~~
2. How does deleting an organization done (Should we limit it in any way - for example only if the organization has no dreams)?
~~3. Who can create / delete an organization? I assume root admin of some kind? @aerugo @skoteskote ~~
~~4. What's the best place to inject subdomain into the app - A short call with @gustavlrsn is needed~~
5. What happens if I don't have an organization and going directly to the root website?

rootux:
@gustavlrsn Since `orgnization` are going to be core - I was thinking about adding it to the `Apollo server` `context` - similar to how we have `currentUser` - so we don't have to get it every time - When we extract `currentUser` (in `api/index.js` we can also extract the organization).
What do you think?

rootux:
Also @gustavlrsn do you have any experience navigating through creating those subdomains?

gustavlrsn:
[...]

PR #63, opened 2020-07-08
Organization

rootux:
# Implements
Solves #58
Solves #81 

# Database action needed
- [x] Locally on dev - Create a root organization called `dev-org` and assign all `events` and all `users` to that organization
- [x] Remove the `events` index `slug` (As now events index uniqueness on slug is slug + organization)
- [x] Remove the `users` index `email` (As now events index uniqueness on email is email + organization)

# Known issues
on the 'create organization` page - entering an empty email doesn't block the user and creates the organization but fails to create the user - should block it on UI

# Not implemented (After discussion)
- Create a backend user interface to allow Deleting an organization
- It's becoming more confusing to start a new env - probably would be easier if we have something in the landing page with existing `organizations`

# Tophatting 🎩
* Created 2 events with the same slug id and different subdomain 
* 2 created users with the same email does not share the same login / record on db
* One organization admin cant do actions for another organization
* Tested full budgeting flow
* Tested creating new dreams
* Tested adding co-creating users
* Cant login if currentorg is not set even for an older token
* An invalid organization url leads to a "404 - Organization not found" text

* Created a new organization and an event under it

* Added `dreams.customdomain` to `/etc/hosts` then set `dreams.customdomain:3000` as custom domain and it worked properly including logging in using a magic link

# Questions
- Currently we only support https for production domains and http for local domains - should we allow it on the `Organization` model - `ssl enabled` or other flag? We use that when we send the magic link email for the domain

# Screenshots
Screen Shot 2020-08-02 at 19 27 52

Screen Shot 2020-08-02 at 20 25 11


rootux:
@gustavlrsn I remember you mentioned an issue with some domains - preventing us from deploying this PR as it would kill production - do you remember what was that about?

rootux:
I believe I've remembered now - for each organization have the ability to state it's domain and by doing so we would redirect based on the domain url.
Are there any security measures we need to verify such as domain validation? I would assume not - as if a user decides to host the website on a specific domain we should allow to load any project state by him. The only requirment is uniquness

rootux:
Testing with custom domain

rootux:
Added a 404 for not found organizations
Screen Shot 2020-07-27 at 16 37 44

rootux:
I've started to design the `Create Orgaznization` page (Which would also be the `Edit organization` page)- 
Currently, it's used internally - but probably in the near future, we would want to give users the ability to create those organizations.

(Please ignore the color scheme as this is only a draft to understand the components)
You can edit it by following this
https://www.figma.com/file/N9O2mfYylNV2ZkWOuyME0R/Material-Baseline-Design-Kit-Community?node-id=4%3A0

Screen Shot 2020-07-28 at 1 17 14

@gustavlrsn what to you think?

aerugo:
> I've started to design the `Create Orgaznization` page (Which would also be the `Edit organization` page)-
> Currently, it's used internally - but probably in the near future, we would want to give users the ability to create those organizations.
> 
> (Please ignore the color scheme as this is only a draft to understand the components)
> You can edit it by following this
> https://www.figma.com/file/N9O2mfYylNV2ZkWOuyME0R/Material-Baseline-Design-Kit-Community?node-id=4%3A0
> 
> Screen Shot 2020-07-28 at 1 17 14
> 
> @gustavlrsn what to you think?

I think this makes a lot of sense. Making this easily available will speed up the adoption rate a lot.

rootux:
Moved all of our users and all of the previous events to the new organization `Blivande` with the `custom domain` - `dreams.blivande.com`

rootux:
The new screen for adding new organizations `/create-organization`
Screen Shot 2020-07-31 at 1 29 28

rootux:
Added ability to see existing organizations under: `root-admin/organizations` - Notice that the user must have the `isRootAdmin` flag set to `true` - This allows us to Delete misconfigured organizations and easily see and manage existing organizations

Screen Shot 2020-08-02 at 20 25 11

No access message
Screen Shot 2020-08-02 at 20 30 11

gustavlrsn:
Great work @rootux! Merging this ✅

Added a fix so we don't send forth the subdomain if you are on a custom domain (it would send the subdomain of the custom domain which we don't care about and could then potentially retrieve the wrong org if there is another org with that subdomain registered).

Also added an optional logo field to the organization model.

PR #75, opened 2020-07-24
Business logic layer

LudwikJaniuk:
GraphQL is nice but should not house the important logic of the application. What if we decide to add REST endpoints in the future, or whatever? So this is a start on implementing a business logic layer that is abstracted from _how_ the backend is accessed. Right now, only the endpoints relevant to grants have been moved, as I'm doing this in preparation for my logging feature. Also, this has resulted in many deduplications and simplifications. Plus, tests for the features I'm moving, to make sure it's done right and hopefully doesn't introduce regressions.


aerugo:
Thanks! We will need to wait with this one until @gustavlrsn is back from vacation, as I think he should be the one reviewing it.

PR #76, opened 2020-07-24
Feat/custom fields position

rootux:
# DB Changes ⚠️
Before merging
- [x] Make sure for each `event` that has `custom-fields`, all `custom-fields` have `position` of type `Double` and set to 1000 then 1001 and so on
**Otherwise the dreams won't load and would show an error**

# Known issues
Since we are using a happy ui - if a user drags and drops then in less then a seconds drags and drops again - the ui would flicker for a moment

# Custom fields
- Supporting position for custom fields (Similar to how trello implemented position of cards)
- Added a new library http://clauderic.github.io/react-sortable-hoc
- Duplicated the custom fields locally to support dragging and drop instant feedback

# Tophatting
Tested moving an item to begin and to the end

Screen Shot 2020-07-24 at 20 09 41

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,9 @@ npm run db:reset
 
 Just visit http://localhost:8081
 
+## Migrations
+Until we have a proper migrations system - make sure you run migrations from `api/migrations` when instructed in the PR
+
 ## License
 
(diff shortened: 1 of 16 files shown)


rootux:
After a call with @aerugo we've decided to merge - let's talk about that once you're back @gustavlrsn

PR #78, opened 2020-07-29
Feat/gal staging

rootux:
Staging of all gals latest additions - `Orgs` and `Position for custom fields`


rootux:
Used for testing

PR #85, opened 2020-08-11
Feat/orgs2

rootux:
# Description
* Ability to edit an organization
* Ability to upload organization image using Cloudinary

# Screenshots
Screen Shot 2020-08-11 at 2 29 33

Screen Shot 2020-08-11 at 17 39 16

Screen Shot 2020-08-11 at 17 35 21

Screen Shot 2020-08-11 at 17 35 30

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ This builds and serves both the API and the UI with one command, and provides ho
 
 # Postrequisites
 After running for the first time add default organization
-By visiting http://dev-org.localhost:3000/create-organization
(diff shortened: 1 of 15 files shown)


gustavlrsn:
Nice work. Did a small change, made the organization logo just a string since we are only using the small image, and the same image was saved for both small/large anyway. Merging!

rootux:
Hmmm. Yes. The issue there is that in the future we might want different
logo sized (Different UI places) - so that case was meant to support for
future versions - but that's okay to use it like sso for now

PR #90, opened 2020-08-14
Fixes #86 - Added tooltip to custom domain

### LINKED ISSUE #86: Added tooltip on organization
Added a Question mark with a popup tooltip (With info on setting the domain) - Near the "Custom Domain" field on Edit / Create Organization

rootux:
Screen Shot 2020-08-14 at 17 44 51


rootux:
@gustavlrsn Are you okay with the UI and the message?
@aerugo Are you okay with the message?

aerugo:
I would make some minor changes to the copy:

You can provide a custom domain here, for example, "dreams.yourdomain.org". Do not add "http://" to your domain.  To enable your custom domain, the CNAME record for that domain needs to be set to yourorgname.dreams.wtf in your DNS settings.

rootux:
Got it.
BTW. They don't need their CNAME to point to the yourorgname.dreams.wtf
it's enough to point to dreams.wtf

PR #96, opened 2020-08-17
Fix edit budget form

gustavlrsn:
Fixes a bug reported on the [Blivande forum](https://forum.blivande.com/t/testing-blivande-dreams/1078/26?u=gustav).

React hook form has a special hook to deal with dynamic array forms, **useFieldArray**, which we were not using and was causing some weird bugs. Using the correct hook solved everything 😃


rootux:
Should we use this in other places where we change dynamic variables - such
as the custom fields?

PR #135, opened 2020-11-17
WIP: Connect to keycloak via auth0/nextjs-auth0 library

gustavlrsn:


gustavlrsn:
Turns out the [@auth0/nextjs-auth0](https://github.com/auth0/nextjs-auth0) library is a standard OIDC implementation and can be used for any OIDC server with Next.js. This solution provides access to the session on the server-side :-)

PR #158, opened 2021-03-09
Add event hub to emit events for integrations

gdpelican:
This is a WIP branch to explore an interface approach to allow for easy integration between the Dreams platform and Discourse (and other integrations as well)

Idea is a pub/sub approach, where each mutation will emit an event which can be subscribed to with a channel name
```
# in subscriber
eventHub.subscribe('create-dream', ({ dream, actor }) => console.log('Dream created, publish to Discourse!')
eventHub.subscribe('create-dream', ({ dream, actor }) => console.log('Dream created, publish to Loomio!')

# in mutation
eventHub.publish('create-dream, { dream, actor: currentOrgMember })
```


gdpelican:
@gustavlrsn Would love your thoughts on this one; this idea to to allow for an easy way for us to publish events from Dreams into other platforms, without bloating the code with lots of if/else branches / 'is this integration enabled?' checks. If it seems like a good approach, I'd pull out the `discourse` calls into the subscribers, and start refining their functionality from there. Let me know!

gustavlrsn:
> @gustavlrsn Would love your thoughts on this one; this idea to to allow for an easy way for us to publish events from Dreams into other platforms, without bloating the code with lots of if/else branches / 'is this integration enabled?' checks. If it seems like a good approach, I'd pull out the `discourse` calls into the subscribers, and start refining their functionality from there. Let me know!

I think it looks great. This will make integrations a lot cleaner, and improve the structure a lot, nice!

gdpelican:
Here's a screenshot of the markdown generated for a discourse post, which gets synced anytime one of those fields changes (NB that I've since updated it to be posted by the current user, rather than 'system')

Captura de Pantalla 2021-03-16 a la(s) 8 41 45 a  m

gdpelican:
The images, currently, are a bit huge. I'll need to see if Discourse supports straight HTML in its markdown

```
### Some title

<img src="image.png" style="width: 200px; float: left" />
```
or has some other scheme for resizing images within posts

gdpelican:
Alright, I've put in a couple fixes here to make this play nice with unpublished dreams, as well as pulling out the MIN_POST_LENGTH into an ENV variable so it can be set per-instance

gustavlrsn:
@gdpelican I've made a few smaller changes including
- adding a link to Dream in the generated markdown (also makes us avoid "Content too short" error when publishing an empty dream
- throwing errors from the discourse API to help debugging
- adding try/catch blocks to prevent crashing app when there is an error
- added minPostLength in the discourse config in the Organization mongoose model (instead of using env variable as every org may have a different setting in discourse) 

## Some remaining issues I'm finding

- Can't add comment to dream that was once published but no longer published
- Can't delete or edit mongodb comments in unpublished dream
- Publishing a dream throws error from discourse API `The requested URL or resource could not be found.` However, on page refresh I can see that the dream was published successfully.
- Publishing a dream that was previously published throws error: `Title has already been used`. 
  - Btw, what should happen to the discourse thread when a dream is unpublished? @aerugo @gdpelican thoughts?

There seems to be some issues with the following section that are in a few places, e.g. the `edit-dream` subscriber. When `discourseTopicId` is missing, we publish the `create-dream` event and then fetch the dream again, but since the publish and subscriber does not wait on eachother the field is still missing when we fetch the dream again.

```js
if (!dream.discourseTopicId) {
  await eventHub.publish('create-dream', { currentOrg, currentOrgMember, event, dream });
  dream = await models.Dream.findOne({ _id: dream.id });
  // dream does not have `discourseTopicId` here as expected by the discourse api call below
}

const post = await discourse(currentOrg.discourse).topics.getSummary({
  id: dream.discourseTopicId
}, {
  username: 'system',
  apiKey: currentOrg.discourse.apiKey,
});
```

> > When adding a comment it is not added in the UI until a page refresh
> 
> This is an interesting one. In an ideal world, posting to Discourse is a side effect and doesn't slow down the resolver at all, and instead once the comment has resolved it notifies a GraphQL subscription so the client gets live updated.
> 
> https://www.apollographql.com/docs/react/data/subscriptions/
> 
> I believe having a subscription has the added benefit of allowing people to post comments on Discourse which will also show up in the Dreams interface automagically.
> 
> However, it doesn't seem like the fastest way to get this up and going, so maybe we just want to chuck an `await` in there for now?

Subscription solution sounds neat but yes maybe not the fastest way to get it to work, although I'm not sure putting an `await` somewhere will work, since there is no awaiting between a publisher and subscriber, right?

api/subscribers/discourse.subscriber.js
```diff
@@ -0,0 +1,218 @@
...
+        eventHub.publish('edit-comment', { currentOrg, currentOrgMember, event, dream, comment });   
+      } catch (err) {
+        console.error(err);
+      }
+    });
+
+    eventHub.subscribe('publish-dream', async ({ currentOrg, currentOrgMember, event, dream }) => {
+      dream.discourseTopicId
+        ? eventHub.publish('create-dream', { currentOrg, currentOrgMember, event, dream })
+        : eventHub.publish('edit-dream', { currentOrg, currentOrgMember, event, dream })
```
gustavlrsn, line 82: Should this conditional be flipped?

aerugo:
> Btw, what should happen to the discourse thread when a dream is unpublished? @aerugo @gdpelican thoughts?

It stays as it is. In the cases where it contains something that should be hidden, that will have to be done by the discourse mods.

aerugo:
Ping @gdpelican, let me know if all of the issues @gustavlrsn raised can be resolved in some way that makes sense.

> Can't add comment to dream that was once published but no longer published

That comments need to somehow work even when the dream is not published is because sometimes issues need to be hashed out between dreamers and admins before publishing. 

As for the issue of what to do with the comments when linked to a Discourse thread, here is my idea:

* Create the topic on Discourse before publishing, add the comments there, but keep the Discourse topic post blank before publishing. That way, text and images don't prematurely make it onto Discourse before publishing on Dreams.
* To make sure that we can keep the link to the topic for the admins that find their way to the dream through Discourse, I think we need to also add a feature that restricts _viewing_ of an unpublished dream to dream creator, co-creator, admins and guides. This probably makes sense to do anyway, and should not be hard to implement.
* Trade-off will be that in the cases where the Dream is unpublished on an instance that uses Discourse, the conversation about the Dream will still be public if you know where to look on the Discourse instance. In the future, this could be improved by having unpublished topic threads in a restricted category, and moved to an unrestricted category when published. For now, this can just be done manually in the very few edge cases where a sensitive discussion is happening on an unpublished dream.

What do you think @gdpelican ?

gdpelican:
Yeah I'm thinking we need to have a discourse topic at all times; dancing around whether a topic exists or not, and having two potential sources of truth for comments  (mongodb and Discourse) at any given time is pretty killer. Discourse has the concept of an 'unlisted' topic which I think is the behaviour we want; I can't recall the _exact_ specs of what unlisted does, but it should be something like 'visible only to admins'

So by default a topic will be posted to Discourse as 'unlisted', and sync all changes (comments, description, etc). Then, publishing (or unpublishing) a dream will simply mark the post as 'listed' or vice-versa

(I'll have a poke at this today)

aerugo:
Yup, that sounds like what we want @gdpelican.
If Discourse is enabled for an event, all comments should go to Discourse.
If not, all comments go to mongodb.

gustavlrsn:
> Yeah I'm thinking we need to have a discourse topic at all times; dancing around whether a topic exists or not, and having two potential sources of truth for comments (mongodb and Discourse) at any given time is pretty killer. Discourse has the concept of an 'unlisted' topic which I think is the behaviour we want; I can't recall the _exact_ specs of what unlisted does, but it should be something like 'visible only to admins'
> 
> So by default a topic will be posted to Discourse as 'unlisted', and sync all changes (comments, description, etc). Then, publishing (or unpublishing) a dream will simply mark the post as 'listed' or vice-versa
> 
> (I'll have a poke at this today)

Really nice with the unlisted/listed functionality when publishing. Really simplifies things a lot and seems to work good on my end. Only thing I think we might miss now before merging is updating the UI when you leave a comment.

gdpelican:
Cool, yep, it feels like a much more solid solution to me too. I've also installed some websockets and GraphQL subscriptions to allow for live updating, as I wasn't able to find a solution that felt good with awaiting the Discourse calls (this is what we want anyway, as it'll mean that everyone gets the updates as they occur)

I have some additional time blocked off tomorrow to get this one up to speed, as you say the big thing missing is the ability for adding new comments to update on the client.

aerugo:
What is the status on this PR now @gdpelican ?

gdpelican:
Need to ensure this is working properly for non-Discourse organizations, but otherwise should be good to go. I'll give an update later today on this one :)

gdpelican:
Alright, I can confirm that adding / editing / deleting comments works with both Discourse and non-Discourse orgs, and that publishing / unpublishing successfully lists and unlists the topic on Discourse. Editing the dream's summary / budget / etc also looks good to go. @gustavlrsn would you like to do a pass on this and see if there's behaviour I'm missing? Specifically I'm a little worried about what's going to happen to existing dreams which have some comments in mongodb and some on Discourse; we may need to run something to publish all of the existing content onto Discourse somehow?

The live update also seems to be work, although it currently is only triggered on Discourse orgs (I can have a look at changing that once this goes in)

gustavlrsn:
> Alright, I can confirm that adding / editing / deleting comments works with both Discourse and non-Discourse orgs, and that publishing / unpublishing successfully lists and unlists the topic on Discourse. Editing the dream's summary / budget / etc also looks good to go. @gustavlrsn would you like to do a pass on this and see if there's behaviour I'm missing? Specifically I'm a little worried about what's going to happen to existing dreams which have some comments in mongodb and some on Discourse; we may need to run something to publish all of the existing content onto Discourse somehow?
> 
> The live update also seems to be work, although it currently is only triggered on Discourse orgs (I can have a look at changing that once this goes in)

Great job! 

I find one behavior that is missing, when adding a discourse config when there are already existing dreams, I can't add new comments. When adding a comment and the dream does not have a discourseTopicId, it would be good to create the thread on Discourse. I tried publishing/unpublishing the dream as a workaround but it errors on unpublish. 

As for old mongodb comments, maybe the quickest way is to merge them with discourse comments in the resolver for comments on dreams? Since we don't have discourse usernames and api keys for every user I think migrating the comments will be too cumbersome.

aerugo:
> As for old mongodb comments, maybe the quickest way is to merge them with discourse comments in the resolver for comments on dreams? Since we don't have discourse usernames and api keys for every user I think migrating the comments will be too cumbersome.

I have pondered this a bit.

Since only Blivande has a Discourse integration, I am making the decision for us to keep things simple and scrap backwards compatibility. 

1) When the Discourse integration is enabled, all comments on mongodb are simply ignored and not shown on Dreams. If the Discourse integration is disabled, the old mongodb comments are shown again. We will just need to make sure that orgs who will need Discourse enable it while onboarding. Later, we could think about some sort of migration logic, but let's not make things too complicated for now.

2) For the Blivande org, I there are only a handful (less than 5) of comment threads that I care about keeping. I am migrating those manually, so we will just need to set the topic ID for those Dreams manually in the database.

aerugo:
These are the only topics that we need to set a Discourse topic for manually. I have migrated the comments and assigned them to the correct users, and in one case I have assigned them to a dummy user.

https://dreams.blivande.com/gallery/5fb0ee70564bcb00217a6702 
Assign: https://forum.blivande.com/t/hot-air/2274

https://dreams.blivande.com/gallery/5fbcc026564bcb00217a67e5
Assign: https://forum.blivande.com/t/that-gift-called-space/2275/8

https://dreams.blivande.com/gallery/5fc3f1d5564bcb00217a6860
Assign: https://forum.blivande.com/t/portrays/2276

https://dreams.blivande.com/gallery/5f907d593c1777002191d5ca
Assign: https://forum.blivande.com/t/we-make-pie/2277

https://dreams.blivande.com/gallery/5fb9054a564bcb00217a671c
Assign: https://forum.blivande.com/t/symbiosis/2278

https://dreams.blivande.com/gallery/5fbc38b0564bcb00217a67d8
Assign: https://forum.blivande.com/t/vill-du-till-himlen/2280

aerugo:
@gustavlrsn, just adding those comment threads to those 6 dreams and then ignoring all mongodb comments should be an easy an safe fix, right?

gustavlrsn:
> @gustavlrsn, just adding those comment threads to those 6 dreams and then ignoring all mongodb comments should be an easy an safe fix, right?



> @gustavlrsn, just adding those comment threads to those 6 dreams and then ignoring all mongodb comments should be an easy an safe fix, right?

Yep, sounds good! I've added those comment threads for those dreams in the database.

gdpelican:
Cool, I've made it so that if no `discourseTopicId` is set on a dream when adding comments / editing / publishing, we first push the dream to Discourse. Any other things I should look at here?

gustavlrsn:
Getting this error when trying to create a new dream: `GraphQL error: Can't save() the same doc multiple times in parallel. Document: 607ea9b301330ec9b7631c18`

gdpelican:
Bleh. Alright I've thrown in an await to make sure we're not saving
multiple things at the same time.

gustavlrsn:
Great, I think it should be good to merge now!

aerugo:
@gdpelican - see it in action! https://forum.blivande.com/t/medicinal-garden/2204/33 :)

aerugo:
Also https://dreams.blivande.com/rrcr/607e784fab7bb0002d1a1113

gdpelican:
Amazing! I'll get started on comment pagination this week :+1:

PR #168, opened 2021-03-22
Add navigation between different org settings

Powersource:
closes https://github.com/Edgeryders-Participio/multi-dreams/issues/163

* [x] fix weird behaviour when pressing enter on general tab


Powersource:
image

image

Powersource:
not the prettiest so far but does the job!

gustavlrsn:
Nice, we can update the styling later, more importantly there is now a way to update org settings without knowing the secret link :)

PR #186, opened 2021-03-28
Add onboarding todo-list on events page

Powersource:
Closes https://github.com/Edgeryders-Participio/multi-dreams/issues/172

* Wait with adding the Realities todo-item, we'll add that once we have better access to the realities api from dreams

* [x] Only show the component if
  * [x] The user is an org admin
  * [x] The user hasn't clicked 'skip'
  * [x] The user hasn't finished all the todos
* [x] Add a progress bar
* [x] Fix text link in first item


Powersource:
progress so far

image

aerugo:
How is this view accessed? If I navigate to the "party event", how do I get
back to the todo list?

Powersource:
This is the normal select-event view, so in this case the url is `http://dev-org.localhost:3000/`

Powersource:
so to get back you click on "dev dreams" in the top left corner

aerugo:
Ah, I see! So the todo-list is only shown to org admins? And it is
dismissed either manually or by completing the tasks?

Powersource:
Yes that's the plan :)

Jacob

Powersource:
image

ui/components/TodoList/index.js
```diff
@@ -0,0 +1,166 @@
...
+const TodoList = () => {
+  const { data, loading, error } = useQuery(GET_TODO_INFO);
+  const [setTodosFinished] = useMutation(SET_TODOS_FINISHED);
+  const [loc, setLoc] = useState("");
+  const [allDone, setAllDone] = useState(false);
+
+  const rawTodos = [
+    {
+      title: "Create community",
+      desc: `This is your own home on the Plato platform, now available under ${loc}`,
```
gustavlrsn, line 74: I think `loc` could be more simple: `${currentOrg.subdomain}.${process.env.DEPLOY_URL}`, to avoid using `useEffect` and relying on the `window` object which is only available on the client side render

Powersource: Yeah thanks that's much more reasonable

Powersource: fixed

PR #189, opened 2021-03-29
Add invite functionality

gustavlrsn:
Ability to invite org members and event members. Right now, it is more of an "add members" functionality.

## Improvements

- [X] Update template for initial email --> "Someone has invited you to Plato"
- [ ] Update Keycloak signup flow pages


gustavlrsn:
I was able to slightly improve the email that is sent when someone is "invited":
image

Still, it is quite anonymous and does not say who invited you, or to what organization or event. For this I think we need to implement a step before or at the same time as the Keycloak signup process..

api/resolvers/index.js
```diff
@@ -1201,41 +1204,185 @@ const resolvers = {
...
-    //   const memberObjs = emailArray.map((email) => ({
-    //     email: email.trim(),
-    //     eventId: currentMember.eventId,
-    //     isApproved: true,
-    //   }));
+      const emails = emailsString.split(",");
 
-    //   let members = [];
+      if (emails.length > 1000)
+        throw new Error("You can only invite 1000 people at a time");
```
Powersource, line 1230: Our current limit on mailgun is 1250 per month :p we could increase that of course but maybe also healthy/to avoid spam, capping it at 100 for now could be good?

api/resolvers/index.js
```diff
@@ -1201,41 +1204,185 @@ const resolvers = {
...
-    //   if (members.length > 0) await sendInviteEmails(members, event);
+        if (user) {
+          const orgMember = await OrgMember.findOne({ userId: user.id });
+          if (orgMember) {
+            const eventMember = await EventMember.findOne({
+              orgMemberId: orgMember.id,
+              eventId,
+            });
+            if (eventMember) {
+              // TODO: use upsert here to adjust isApproved
```
Powersource, line 1248: todo now?

aerugo:
We will definitely need to go higher than 1250 per month pretty soon though.

api/resolvers/index.js
gustavlrsn: Fixed this

api/resolvers/index.js
gustavlrsn: I think Kiezburn had around 600 people to add, so would be nice to allow them to do it once I guess

PR #196, opened 2021-04-09
Improved granting/funding system

gustavlrsn:


gustavlrsn:
Current status of event members table with allocate funds modal:

image

gustavlrsn:
Summary of new granting functionality
- got rid of tokens/grantlings, grant value, total budget, grants per member, max tokens to one dream per user (but added max amount to one dream per user)
- replaced with using regular currency as the basis for granting and funding goals
- all amounts are stored as integers representing cents of whatever currency the event has chosen, this is to avoid floating point precision problems (i.e. `0.1 + 0.2 == 0.3` is `false` in javascript)
- admins can allocate money to event members (Allocation model in db)
- event members can contribute money to dreams (Contribution model in db)
- dreams automatically get status `funded` when they reach their max funding goal
- dreams can also accept funding when their minimum goal is met (for events that allow stretch goals)
- dreams can cancel funding by dream creators or event admin, then the event members will have their contributions back to be spent again
- dreams can be marked completed, then funding can't be canceled
- dream creation closes, granting opens and granting closes are optional dates. if not set you can do whatever whenever

PR #199, opened 2021-04-11
Don't break when an org doesn't have discourse

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/198


Powersource:
I'm just gonna merge this @gustavlrsn since stuff is buggy atm

PR #206, opened 2021-04-18
Update to apollo client 3

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/203

* [x] Fix readonly error on dragging sortable fields
* [x] Fix page not loading when visiting directly/reloading something other than /


Powersource:
This seems to work well except for somehow causing a bunch of this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Read-only which at least breaks the app in dev

PR #216, opened 2021-04-20
Add custom fields to discourse post

aerugo:
Adds custom fields to Discourse version of Dream. 
Perhaps there is a smarter way to do this that is more in line with our standards. Let me know if that's the case.


gustavlrsn:
Looks perfect to me

PR #228, opened 2021-04-23
Start moving over basic parts of the realities UI

Powersource:
Url format should be similar to the current format on the old realities app when this pr is done (but with `/realities/` prepended to the path). There's a risk react/next throws out too much state when navigating between pages, like e.g. broken animations, but we'll see how it goes because pretty urls would be nice. If that doesn't work great we'll later move to something like `/realities/?respid=asdf` so that we can do [Shallow routing](https://nextjs.org/docs/routing/shallow-routing) in next

* move over functionality so people can use it, then merge this. we don't need a pretty ui yet
* skip "fullscreen" functionality


Powersource:
Getting there slowly. Moving this directly to material-ui as well because importing the bootstrap css broke the rest of the app.

image

aerugo:
Looking good @Powersource !

Powersource:
1 sec gonna add search on this branch anyway

gustavlrsn:
Looks good to me. I have not been able to test it locally, got the Cors error that seems like it should not be there from the setup in realities.. and then found out that Neo4j does not run on M1 macs. But since this is all contained to a non-visible page I think we can merge it

PR #231, opened 2021-04-27
Add footer with gdpr links

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/208

image

image


skoteskote:
It's not really necessary to have ToS in the footer, as that's what the person signing up is agreeing to. But guess it doesn't hurt either :)

Powersource:
Yeah I was thinking we might as well put it there :)

PR #232, opened 2021-04-27
Bulk allocation of funds and ability to set (instead of add) balances

gustavlrsn:
Adds ability on the event members page to bulk allocate funds to members, both adding and setting. Also adds the Set option for individual allocations.

### DIFF
--- a/api/resolvers/index.js
+++ b/api/resolvers/index.js
@@ -1662,8 +1662,8 @@ const resolvers = {
     },
     bulkAllocate: async (
       _,
-      { eventId, amount },
-      { currentOrgMember, models: { EventMember, Allocation } }
+      { eventId, amount, type },
+      { currentOrgMember, models: { EventMember, Allocation, Contribution } }
     ) => {
       if (!currentOrgMember) throw new Error("You need to be logged in.");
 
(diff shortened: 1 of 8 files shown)


ui/components/members/AllocateModal.js
```diff
@@ -34,14 +40,20 @@ const AllocateModal = ({ member, event, handleClose }) => {
...
+          selected={type}
+          className="mx-auto mb-4"
+        />
+        {/* <p className="text-center mt-4 mb-2">
           <span>
             {thousandSeparator(member.balance / 100)} {event.currency}{" "}
           </span>{" "}
           +
-        </p>
+        </p> */}
```
Powersource, line 56: remove

ui/components/members/AllocateModal.js
```diff
@@ -61,9 +73,23 @@ const AllocateModal = ({ member, event, handleClose }) => {
             autoFocus
             className="w-36 mx-auto mb-2"
           />
-          <p className="text-center mb-4">
-            = {thousandSeparator(total / 100)} {event.currency} total
+          <p className="text-center mb-4 text-sm text-gray-800">
+            {type === "Add" ? (
+              <>
+                Adding {thousandSeparator(amount / 100)} {event.currency} to{" "}
```
Powersource, line 79: Maybe rename `amount` to cents or something? Idk not super needed but I started mixing up all the different variable names when I got to this part :p

ui/components/members/BulkAllocateModal.js
```diff
@@ -0,0 +1,92 @@
+import { useState } from "react";
```
Powersource, line 1: Could maybe be nice to refactor this file and AllocateModal to deduplicate code in some way. But it's not a ton of duplication so only if you feel like it

api/resolvers/index.js
```diff
@@ -1651,15 +1651,116 @@ const resolvers = {
...
+
+        const adjustment = amount - balance;
+
+        await new Allocation({
+          organizationId: currentOrgMember.organizationId,
+          eventId: targetEventMember.eventId,
+          eventMemberId,
+          amount: adjustment,
+        }).save();
+      }
```
Powersource, line 1694: Could be nice with an `} else {` to catch if you send a bad type (or does the graphql enum maybe check that for you?)

api/resolvers/index.js
```diff
@@ -1651,15 +1651,116 @@ const resolvers = {
...
+
+          const adjustment = amount - balance;
+
+          await new Allocation({
+            organizationId: currentOrgMember.organizationId,
+            eventId,
+            eventMemberId: member.id,
+            amount: adjustment,
+          }).save();
+        }
```
Powersource, line 1759: This chunk of code inside the for loop looks like it's mostly the same as the code in the other `allocate` resolver. Would be nice to extract the common logic into a shared function.

api/resolvers/index.js
gustavlrsn: Yep, the graphql enum will take care of that

ui/components/members/AllocateModal.js
gustavlrsn: fixed

ui/components/members/BulkAllocateModal.js
gustavlrsn: Let's go for rule of three, "Three strikes and you refactor" :-)

api/resolvers/index.js
gustavlrsn: I consolidated into a function that I put i `api/controller/index.js` where we can add more controller type functionality, inspired by #75

ui/components/members/AllocateModal.js
gustavlrsn: All amounts are in cents so renaming to that might just make it more confusing 🤔

Powersource: Yeah good point. Maybe just `amount` that's in full units

api/resolvers/index.js
Powersource: Beautiful!

PR #235, opened 2021-05-03
Require the user to accept the ToS and privacy policy

Powersource:
fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/207

This PR doesn't actually do the change itself, but it adds the script that i used to do the change. We'll be able to use it (a slight modification of it) if we need to ask users in the future to accept the ToS again.

This was just for already existing users. The normal setting in keycloak will make sure new users need to accept the ToS

### DIFF
--- a/api/scripts/requireAcceptToS.js
+++ b/api/scripts/requireAcceptToS.js
@@ -1,3 +1,6 @@
+const {
+  RequiredActionAlias,
+} = require("keycloak-admin/lib/defs/requiredActionProviderRepresentation");
 require("dotenv").config({ path: "api/.env" });
 const initKcAdminClient = require("../utils/initKcAdminClient");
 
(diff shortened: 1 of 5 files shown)


api/scripts/requireAcceptToS.js
```diff
@@ -0,0 +1,56 @@
...
+  RequiredActionAlias,
+} = require("keycloak-admin/lib/defs/requiredActionProviderRepresentation");
+require("dotenv").config({ path: "api/.env" });
+const initKcAdminClient = require("../utils/initKcAdminClient");
+
+if (!process.env.KEYCLOAK_AUTH_SERVER) {
+  throw ".env not loaded. Call from the top dir in the repo";
+}
+
+const TERMS = RequiredActionAlias.terms_and_conditions;
```
gustavlrsn, line 11: We should also add this required action in the `inviteEventMembers` and `inviteOrgMembers` resolvers

Powersource: I believe that would probably be taken care of by the setting in keycloak that requires the ToS to be accepted by new users (that I've enabled). But maybe we need to test that.

gustavlrsn: Tested it now, it becomes part of the "Complete signup" process as the last step. It's not listed as one of the steps on the first screen shown after clicking that link though

Powersource: Ok good. That's how the ToS accepting works in keycloak by default. You accept it on the screen after registering/loggin in. It's a bit unusual but works. I think I read you could technically maybe hack in a checkbox on the registering screen but I don't know if that would be worth the effort.

Can I mark this as resolved?

Issue #211, opened 2021-04-19
Comment pagination and sorting


aerugo:
We should paginate comments and allow sorting by most recent on top or on the bottom.

This would probably also fix #210

aerugo:
@gdpelican, could you have a look at this once we've finalized #158?

aerugo:
@gdpelican we talked about this in our meeting today. We should implement it so that the last page of comments is the one that is loaded first, and that there is an option to "load more comments above" or similar.

aerugo:
@gdpelican - have you had a chance to look at this yet?

gdpelican:
I have gotten a start, but nothing to show just yet. I have Tuesday this
week blocked out to get this to presentable :)

gdpelican:
Alright, this one is good for a look _with the exception_ of deleting not updating properly; I have time tomorrow blocked off for that one. (Lots and lots of cache invalidation, haha)

aerugo:
Done. Closed in issue clean-up.

PR #241, opened 2021-05-04
Adding comment pagination

gdpelican:
Functionality for #211 

This adds the ability to load backwards from the latest comments in the thread. It's currently set to 3 comments for development; I'll be sure to update it to a more sane value before it's merged.

Currently the only issue is properly updating the cache once a record has been deleted; this should be do-able by using an `update` method on the delete mutation, and the live update should no longer return deleted records so it stays gone :)

NB Once this is in we'll want to circle back and do a better job of paginating the records we get back from Discourse, rather than loading them up each time and windowing them.

* [x] Fix publish/unpublish bug
* [x] Style Load more button
* [x] Add `update` for cache on deletion. i think it's still not there but the cache seems to update properly anyway
* [x] Check that deletion works properly, and check having 2 tabs open at once. work at least with mongo, but isn't updated with subscriptions, see below
* [x] Fix subscription bug with orgMember. no gave up
* [x] Update apollo-server (in a different pr) to see if it helps with the above bug
* [x] Check if subscriptions even work on master. If they don't, stop trying to fix it here
  * Comment subscriptions seem broken on master as well (maybe just in a different way though) so i'll skip fixing them here. moved to https://github.com/Edgeryders-Participio/multi-dreams/issues/314
* [x] Check that the discourse integration seems to work
  * [x] Fix bug when using discourse that makes every commenter look like the current user
  * [x] Check that discourse deletion works
  * [x] Don't show comments in progress of being deleted or at least don't let the user try to delete them again
  * [x] Comments sometimes show up on the wrong dream
  * [x] frihamnstorget not displaying comments for some reason. kulturbryggan does
  * [x] display comments even when a user doesn't have an orgMember (i.e. just posted on discourse)
  * [x] don't display empty system comments (or is there a better criteria?)
  * [x] display system log comments, with the right user icon and name


aerugo:
@gdpelican, this is still WIP and not ready for review yet, right?

gdpelican:
Yes, still WIP on this one. Still some trickiness around the Discourse API
here.

gustavlrsn:
Great stuff. Is it possible to add back total number of comments to the dreams query so that we can display it on the event/collection page?

gdpelican:
I'm a bit worried about that one for Discourse dreams because it involves
an API query, and if we're doing that API query for every dream in an event
it'll be quite slow (The mongodb one is an N+1 query but shouldn't be that
bad).

Maybe we cache the number on the Dreams site so we don't need to go and
fetch it every time?

aerugo:
It’s also fine for us to not have the number of comments for Discourse
Dreams. I think we are adding too much complexity for a use case that is
not going to be used by most orgs. Not having number of comments on the
front page when you have discourse activated is a fine trade off.

gdpelican:
Alright, I've made some good progress here and would love some eyes on this. @aerugo @gustavlrsn 

I've ended up pulling out quite a bit of functionality to the subscribers, including a new Mongo subscriber which handles interacting with the local database (more of this could happen in the future). This means that we now have completely separate paths for what happens in Mongodb vs Discourse.

Now you're also able to get back results from the eventHub, in the format:

```
const { discourse, mongodb } = eventHub.publish("create-comment", { ... })
// {
//   discourse: { something: 'happened on discourse' }
//   mongodb: { something: 'happened in mongodb' }
//   email: { some: 'emails sent' }]]
// }
```
^^
This helps a lot when responding to Graphql mutations synchronously. I believe live update will now work as well across clients, although haven't tested that extensively yet.

Now each Dream page will maintain a CommentSet, which should have a total, from, per, and order. I have it hard-coded to 3 comments per page in development at the moment, but we'll obviously want to bump that up (maybe into a configurable ENV variable or something) once it seems good to go.

I have been having a bit of trouble with Keycloak during the course of this; I think the long response times from NZ may be a culprit? Everyone once in a while I'll get a timeout and a gnarly 401 forbidden error that I've been unable to diagnose; it seems to happen very frequently for a while then works perfectly :/

gustavlrsn:
Regarding the 401, could have been related to an initialization of the keycloak admin client in the eventhub subscribers file (the auth to Keycloak timed out), that is now fixed on `master`.

Powersource:
Just got this when I clicked Publish on a dream. Also on unpublish. Seems to not get it in production. The dream seems to be published anyway at least.

image

aerugo:
@gdpelican could you have a look at that error?

aerugo:
@Powersource, if @gdpelican does not have time to finish this now, this would be a good one to take a look at after you are done with your current task.

ui/components/Dream/Comments/index.js
```diff
@@ -1,46 +1,55 @@
...
 import AddComment from "./AddComment";
 import Comment from "./Comment";
 import Log from "./Log";
+import HappySpinner from "components/HappySpinner";
+import Context, { useCommentContext } from "contexts/comment";
+
+const Comments = ({ currentOrgMember, currentOrg, dream, event }) => {
+  const context = useCommentContext({
+    from: 0,
+    limit: 3,
```
Powersource, line 11: todo: increase

Powersource:
The publish/unpublish bug seems to be the same kind of bug as this https://github.com/Edgeryders-Participio/multi-dreams/issues/198

Powersource:
Having a really weird bug with the subscriptions. The subscription is set up properly and delivers events to the client, but I keep getting an error like this

```json
{
  "type": "data",
  "id": "1",
  "payload": {
    "errors": [
      {
        "message": "Cannot read property 'OrgMember' of undefined",
        "locations": [
          {
            "line": 9,
            "column": 7
          }
        ],
        "path": [
          "commentsChanged",
          "comment",
          "orgMember"
        ]
      }
    ],
    "data": {
      "commentsChanged": {
        "action": "created",
        "comment": {
          "id": "60ddf5948222788950051fe3",
          "content": "test2",
          "htmlContent": null,
          "createdAt": 1625159060472,
          "orgMember": null,
          "__typename": "Comment"
        },
        "__typename": "CommentAction"
      }
    }
  }
}
```

The normal graphql flow seems to work correctly, in that it sees that the orgMember is missing at first so then it uses the resolver for that to get it, but in the end once it gets to the client the field is still null somehow. And if I instead get the orgMember myself and insert it like below, it still turns out as null in the client. Really confusing.

```diff
diff --git a/api/subscribers/mongodb.subscriber.js b/api/subscribers/mongodb.subscriber.js
index 130848d..e7df47c 100644
--- a/api/subscribers/mongodb.subscriber.js
+++ b/api/subscribers/mongodb.subscriber.js
@@ -36,7 +36,16 @@ module.exports = {
         dream.save();
 
         const created = dream.comments.reverse()[0];
-        liveUpdate.publish("commentsChanged", { commentsChanged: { comment: created, action: 'created' } });
+
+        if (created.authorId) {
+          created.orgMember = await models.OrgMember.findOne({
+            _id: created.authorId,
+          });
+        }
+
+        liveUpdate.publish("commentsChanged", {
+          commentsChanged: { comment: created, action: "created" },
+        });
 
         return created;
       }
```

Powersource:
Updated apollo server, didn't help

Powersource:
This seems to be incorrect

> The normal graphql flow seems to work correctly, in that it sees that the orgMember is missing at first so then it uses the resolver for that to get it

The resolver for the author is ran once, but that's for the regular addComment mutation, not for the subscriptions (which there are 2 of in my case with 2 tabs).

aerugo:
@gdpelican - what's your take on this? We weren't able to resolve this with @Powersource 
Did you also hit a wall with this problem?

Powersource:
I'm happy enough with this PR now. I think in general though in the future we should avoid PRs getting this big/complex. I suggest we merge this unless Gustav finds a new bug when testing it. And then we'll improve on this functionality in other PRs.

gustavlrsn:
Seems to be working on my end.

PR #250, opened 2021-05-16
Add tags

gustavlrsn:
Adding lowercased (slugified) strings in a tags array on the dream model.

- ability to edit tags on dream page
- ability to filter on tags on dream collections

### DIFF
--- a/api/resolvers/index.js
+++ b/api/resolvers/index.js
@@ -117,7 +117,7 @@ const resolvers = {
     },
     dreams: async (
       parent,
-      { eventSlug, textSearchTerm },
+      { eventSlug, textSearchTerm, tags },
       { currentOrgMember, currentOrg, models: { Event, Dream, EventMember } }
     ) => {
       let currentEventMember;
@@ -134,7 +134,14 @@ const resolvers = {
         });
       }
 
-      // if admin or guide, show all dreams (published or unpublished)
+      const tagQuery = {
+        ...(tags
+          ? {
+              tags: { $all: tags },
+            }
+          : null),
+      };
+
       if (
         currentEventMember &&
         (currentEventMember.isAdmin || currentEventMember.isGuide)
@@ -142,7 +149,8 @@ const resolvers = {
         return Dream.find({
           eventId: event.id,
           ...(textSearchTerm && { $text: { $search: textSearchTerm } }),
-        });
+          ...tagQuery,
+        }).sort({ createdAt: -1 });
       }
 
       // todo: create appropriate index for this query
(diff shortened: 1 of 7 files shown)


api/resolvers/index.js
```diff
@@ -117,7 +117,7 @@ const resolvers = {
     },
     dreams: async (
       parent,
-      { eventSlug, textSearchTerm },
+      { eventSlug, textSearchTerm, tags },
```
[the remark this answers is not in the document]
Powersource: Yeah I get this error in the browser console when having not-hashtagged text in the search when searching "Error: text index required for $text query"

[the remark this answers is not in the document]
gustavlrsn: Seems like there is some error in creating the indexes in development, unrelated to this PR though and the text index exists in the production db

ui/components/Dream/Sidebar.js
```diff
@@ -105,10 +123,10 @@ const DreamSidebar = ({ dream, event, currentOrgMember, canEdit }) => {
       },
     ],
   });
-
-  const [contributeModalOpen, setContributeModalOpen] = useState(false);
-  const [cocreatorModalOpen, setCocreatorModalOpen] = useState(false);
-  const [actionsDropdownOpen, setActionsDropdownOpen] = useState(false);
+  const [editTags] = useMutation(EDIT_TAGS_MUTATION, {
```
Powersource, line 126: Right now if i add/remove a tag here and go back to the collection page and search, the results will be old. I think the fix would be to add an `update` to the cache like here https://github.com/Edgeryders-Participio/multi-dreams/blob/cf711d187b9da0d9d9effe6e2efc7de8d74a6cc1/ui/components/Realities/index.js#L52

gustavlrsn: This will only occur if you go back to an old query for a tag on the collection page without refresh, if you go back and search again it seems to work on my end?

It'd be a pretty tricky to implement an update cache function for this since you would try recreate the results mongodb gives back when querying for specific tags. Removing an item when there are certain tags and adding an item to the cache when there are certain tags, better to let the API handle that.

Powersource: Hmm yeah maybe a better solution is to make the search query trust the cache less? Maybe the cache-and-network policy? https://www.apollographql.com/docs/react/data/queries/#cache-and-network

gustavlrsn: Pushed a fix that sets 'cache-and-network' policy on the dreams query

PR #252, opened 2021-05-17
Feat/set custom field maxchar

### LINKED ISSUE #92: Add character limit to custom fields
Allow a character limit to be set for custom fields. 

Also, consider allowing surpassing character limit while writing but do not allow saving unless within the limit. This makes it possible to use Dreams as tool for writing and editing.

aerugo:
This PR introduces a number of changes to custom fields:

- Custom fields are now simply called "Questions" in the UI
- Questions now have custom character limits which are displayed in the front end.
- Question custom character limit is set through question edit modal, minor CSS updates to accommodate this
- Increases default size of multiline text field edit box
- Adds markdown formating to questions
- Description is no longer a "default" field. It still exists in the code, but if a Dream does not have it, it is not displayed.
- New Collections automatically get a Question on creation called Description

Closes #92

### DIFF
--- a/api/resolvers/index.js
+++ b/api/resolvers/index.js
@@ -384,6 +384,16 @@ const resolvers = {
         currency,
         registrationPolicy,
         organizationId: currentOrg.id,
+        customFields: [
+          {
+            name: "Description",
+            description: "Describe your Dream",
+            type: "MULTILINE_TEXT",
+            isRequired: false,
+            isShownOnFrontPage: false,
+            position: 1001,
+          },
+        ],
       }).save();
 
       await new EventMember({
@@ -392,19 +402,6 @@ const resolvers = {
         isAdmin: true,
         isApproved: true,
       }).save();
-
-      customField = {
-        name: "Description",
-        description: "Describe your Dream",
-        type: "MULTILINE_TEXT",
-        isRequired: false,
-        isShownOnFrontPage: false,
-      }
-
-      const position = 1001
-
-      event.customFields.push({ ...customField, position });
-      event.save()
 
       await eventHub.publish("create-event", {
         currentOrg,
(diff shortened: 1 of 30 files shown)


gustavlrsn:
I think it looks good! One detail would be to fix markdown support for custom fields (at least for LONG_TEXT), so that descriptions can have markdown in the future as well. Could do that in this PR or create an issue for it, up to you.

api/resolvers/index.js
```diff
@@ -393,6 +393,19 @@ const resolvers = {
         isApproved: true,
       }).save();
 
+      customField = {
```
gustavlrsn, line 396: This is an undefined variable, should have `const` in front of it

gustavlrsn: You could add this object in the first `new Event` call on line 379, like this:

```javascript
  const event = await new Event({
        slug,
        title,
        description,
        summary,
        currency,
        registrationPolicy,
        organizationId: currentOrg.id,
        customFields: [
          {
            name: "Description",
            description: "Describe your Dream",
            type: "MULTILINE_TEXT",
            isRequired: false,
            isShownOnFrontPage: false,
            position: 1001,
          },
        ],
      }).save();
```

and avoid having to save the event twice

gustavlrsn (changes requested):

ui/components/EventSettings/CustomFields/index.js
```diff
@@ -12,9 +12,10 @@ const CustomFields = ({ event }) => {
 
   return (
     <div className="px-6">
-      <h1 className="text-2xl font-semibold mb-2">Custom fields</h1>
+      <h1 className="text-2xl font-semibold mb-2">Questions</h1>
```
gustavlrsn, line 15: Should also update this word in settings menu list

aerugo:
Alright, @gustavlrsn this should be good to go now, including Markdown!

PR #256, opened 2021-05-19
Enable styling keycloak with react

Powersource:
Related to https://github.com/Edgeryders-Participio/multi-dreams/issues/253


Powersource:
In the deployment folder i.e. `/opt/jboss/keycloak/standalone/deployments` in the container there was a readme file with pretty much the contents of (part of) this webpage https://docs.jboss.org/author/display/AS7/Application%20deployment.html Some info on how the deployment works there.

I've managed to deploy the theme but I can't seem to update it for some reason. It's loaded the jar file into it's system but once that's there it doesn't seem to want to be overwritten. Will have to look at that more.

Powersource:
For more context, a discussion on having keycloakify help us style emails too https://github.com/InseeFrLab/keycloakify/issues/9 tldr: not yet, but if we want a simple version of if we could probably hack that together

PR #258, opened 2021-05-20
WIP: Email notifications

gustavlrsn:


gustavlrsn:
I've set this up to send an email on new comments to all cocreators. We discussed not sending to those users that have a discourse account connected, but since new dream posts are created by the 'system' user they won't get any notifications unless they participate in the thread.

api/services/EmailService/email.service.js
```diff
@@ -7,6 +7,71 @@ const mailgun = require("mailgun-js")({
...
+
+    const domain = currentOrg.customDomain
+      ? `https://${currentOrg.customDomain}`
+      : `https://${currentOrg.subdomain}.${process.env.DEPLOY_URL}`;
+    const link = `${domain}/${event.slug}/${dream.id}`;
+    const subject = `${name} commented on ${dream.title}`;
+    const text = `"${comment.content}"\n\nGo here to reply: ${link}`;
+
+    if (emails.length) await this.sendEmail(emails, subject, text);
+    console.log("0 emails sent");
```
Powersource, line 48: forgot a return on the line before?

Powersource:
I didn't test it, but apart from a small comment it LGTM :)

api/services/EmailService/email.service.js
gustavlrsn: good catch 👍 , fixed this

PR #266, opened 2021-06-01
Use redirects to sent people to the terms and conditions and privacy policy

Powersource:
Due to https://github.com/Edgeryders-Participio/plato-core/issues/7#issuecomment-852058636


Powersource:
I'll merge this @gustavlrsn since it's a small change and I need it to apply to google's thing

PR #268, opened 2021-06-04
Move the Privacy Policy, and Terms and Condition to iframes

Powersource:
Contributes to https://github.com/Edgeryders-Participio/plato-core/issues/7 since google seems very bothered by us having our policies on someone else's website


Powersource:
Mergin myself since it's a pretty small change and I want to kick the approval process on again

PR #273, opened 2021-06-11
Paginate dreams in collection page

Powersource:
Progresses https://github.com/Edgeryders-Participio/multi-dreams/issues/191

I'm thinking I'll just add pagination to collections first, and if there's no complaints on the pattern I used I'll go do the other ones in another PR.

Worth comparing with how James has done it for comments in https://github.com/Edgeryders-Participio/multi-dreams/pull/241 , we use a bit different patterns but I think that makes sense since we're getting things from different db's.

* [x] Fix logged out people not seeing dreams


gustavlrsn:
LGTM!

Found one tiny thing that we could live with but when clicking load more until you get all results, and then going into a dream and going back, the Load more button shows up again although all results are displayed.

Powersource:
I'll take the LGTM to mean I'm free to merge :)

Hmm interesting bug. What makes it weirder is that I can't reproduce it in the transaction list where I've just added pagination in pretty much the same way https://github.com/Edgeryders-Participio/multi-dreams/pull/276

Powersource:
hmm i think it might be related to the "cache-and-network" fetch policy on the query. when you go back to the page there's a new query fired for some reason, and that query has offset: 0 (that's the default) which then returns moreExist: true

PR #280, opened 2021-06-19
Feat/show existing funding in progress

### LINKED ISSUE #259: Calculate progress from existing funding + contributions
I think we might want to calculate funding progress including existing funding. I notice that the way we do it now is actually more confusing. Take [this Dream](https://dreams.blivande.com/frihamnstorget-development/5ef2369242d46b0007fddcaa) in the Frihamnstorget collection. It was uploaded when there was no funding to be distributed. Eventually, they just built it with their own money. But it still looks as if it was never funded or built.

If we calculate funding progress including existing funding, that is easy to fix, and the dream shows as funded/completed in the collection. 

Example:
Screenshot 2021-05-20 at 20 47 33

What do you think? I have made this change on a branch to test it: https://github.com/Edgeryders-Participio/multi-dreams/tree/feat/show-existing-funding-in-progress
However, I am not sure if it also has to be changed in other places to be complete.

aerugo:
Fixes #259


gustavlrsn:
Looks good to me, but I'd add back the `package-lock.json` files that should be checked in

PR #300, opened 2021-06-23
Change dreams shuffle over time

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/298

* [ ] Refactor mongodb query to be less lisp
* [ ] Move the time seed to be generated in the ui on page load, so the offset doesn't change while you're on a page, breaking pagination

### DIFF
--- a/api/resolvers/index.js
+++ b/api/resolvers/index.js
@@ -121,6 +121,7 @@ const resolvers = {
       {
         currentOrgMember,
         currentOrg,
+        kauth,
         models: { Event, Dream, EventMember, Tag },
       }
     ) => {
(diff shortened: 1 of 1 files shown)


aerugo:
Is this ready to merge @gustavlrsn and @Powersource ?

Powersource:
It works like this, but atm the way it's implemented lowers ux slightly in other ways, mainly that it doesn't ssr the list anymore. But that's probably a worthy tradeoff for now?

aerugo:
"it doesn't ssr the list" - what does that mean?

Powersource:
That if you follow a link to the collection, you won't get the page with the list of dreams already there, the list will load with a spinner. Tbf this also already happens if you navigate to the event from another page in the app.

Another change would be that if you're on the collection page, navigate somewhere else, and navigate back to it again, the list will load again.

I have an idea for a fix that might work, but imo that can wait until another pr.

aerugo:
Yeah, that's okay, we can live with that for now.

gustavlrsn:
I really think we should aim for SSR to work, but I'm fine with merging this before that if there is priority on getting the time shuffle in place.

One thing we should fix is the loading spinner though, right now only "No dreams..." is displayed when loading

aerugo:
There is not really an urgent need to get the time shuffling working.
Kiezburn have soon finished their funding round.
Gustav, I think we should go with your gut feeling on that SSR is important
for UX and wait with merging this until we have fixed that.

PR #304, opened 2021-06-27
Allow contributing cents

Powersource:
Resolves https://github.com/Edgeryders-Participio/multi-dreams/issues/299

Seems almost too simple, but haven't run into any issues in my (not very elaborate tbh) testing


Powersource:
Makes the up and down arrows next to the input less useful, since they only increment by 0.01 now. But that ux should be able to be improved in potential future PRs. And it's debatable how useful they are even when they increment by 1.

Issue #263, opened 2021-05-25
Users should not be able to comment on dreams in collections they are not a part of


skoteskote:
Currently, a user in an org can comment on any dream in any collection of that org, even if they are not a part of that collection. A user should only be able to comment on dreams in collections they are added to. 

Example: 
* Organisation Blivande has an org with 11 different collections.
* One of these collections is Beta Beautification Program which is invite-only as it's only for current Beta members. 
* Another of these collections is Frihamnstorget2021, which is open registration policy as it's intended for anyone in town. 
* Bengt becomes a member of Frihamnstorget2021 to help out with gardening, they love commenting on other projects which is great in the context of Frihamnstorget since we want to spur engagement.
* Bengt then goes into the Beta Beautification Program and starts commenting on everything, creating noise and confusion since he is not a member of Beta.

PR #318, opened 2021-07-31
fix: user should be a collection member to post comments

hugocortes:
Fixes Issue #263 

API throws an error if a user attempts to add a comment to a dream within a collection they are not a member of. UI does not display AddComment component if they are not collection member.

Screenshot:
Screenshot from 2021-07-31 03-47-51


gustavlrsn:
Great, thank you @hugocortes!

PR #333, opened 2021-09-13
Add basic e2e testing that launches the app and logs in

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/326

Inspiration https://github.com/Edgeryders-Participio/realities/pull/206

* [x] Add the new env vars to production before merging. Ok doesn't look like this is needed


Powersource:
I think making the app work well in the test environment might be a bit tricky, we'll see, since we have to make it interact correctly and safely with the prod keycloak server. I'm thinking I'll do that in a PR that builds on this one.

PR #336, opened 2021-09-18
Fix small mobile ui bugs

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/243


Powersource:
made the org icon never shrink but made the org name disappear on small screens if the org has an icon

PR #337, opened 2021-09-22
Fix e2e tests

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/326 hopefully without breaking anything this time

* [x] Test normal dev build against local/prod dbs
* [x] Test prod build against the dbs
* [x] Doublecheck that prod has all the env vars
* [x] Can we skip having the env var stuff in next.config.js ? Yes I think we can but that's a sidequest https://github.com/Edgeryders-Participio/multi-dreams/issues/338
* [x] Do we actually check that the user gets logged in?
* [x] Make the test check in the top right that the user gets logged in
* [x] Before merging, disable saving videos on success
* [x] Uncomment kcadminclient and fix it so tests pass and the app works
* [x] Update formatError so we don't accidentally leak errors containing kc admin pass in ci

I think the real bug in the previous PR was that we started using KEYCLOAK_REALM in the api but the prod server didn't have that defined. So merging this (with kcadminclient uncommented) should work in prod. But now I'm fixing the tests so they work too.


Powersource:
> Do we actually check that the user gets logged in?

We check that the user is able to fill in their username/password, and also that they after that end up back in the app. I don't think there's a way to do that without successfully logging in. Well, maybe keycloak thinks they're logged in but the app can't contact keycloak for some reason and check that the user is logged in, I think we've had that happen before I think.

Powersource:
Checked a recording of a test and it runs too fast to see if the user actually gets logged in, gonna have to add a check in the test

Powersource:
we decided to try to merge this directly into https://github.com/Edgeryders-Participio/multi-dreams/pull/321

package.json
```diff
@@ -9,11 +9,22 @@
     "storybook": "cd ui && start-storybook -p 6006",
     "postinstall": "cd ui && npm i && cd ../api && npm i",
     "lint": "prettier --ignore-path .gitignore --check . && eslint .",
-    "fix": "prettier --ignore-path .gitignore --write . && eslint . --fix"
+    "fix": "prettier --ignore-path .gitignore --write . && eslint . --fix",
+    "test:run:mongo": "sudo docker rm -fv mongo-dreams-test ; sudo docker run --name mongo-dreams-test --publish=28017:27017 mongo:4.4.3",
```
Powersource, line 13: We should remove this line obviously but from when I wrote this for realities I remember that the whole `rm` thing was important to not end up with tons of leftover docker containers or something, it eats up your storage space otherwise unless you manually run commands to clean up stuff. Maybe not a big problem if we'll mostly run the tests in ci though

Powersource: So wondering if there's a simple equivalent for docker-compose

Powersource:
Successfully launches the app and clicks login. Doesn't get further than that because idk how to get login to work yet :P if it's even possible so far

package.json
Powersource: Maybe this from a few lines above is a good hint 

```
"db:up": "docker-compose up",
"db:reset": "docker-compose down && docker-compose rm -f",
```

Powersource: I think for now it"s fine to not have a command to clean it since we're using the same db instance for normal development. We'll handle this in the future when we create a new db to be able to wipe it afterwards

PR #341, opened 2021-09-28
Update budget ux/copy

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/340 and a bit more

* [x] Changing from "Expenses" to "Costs"
* [x] Adding a text line on that "your funding target is automatically calculated through costs-existing funds"
* [x] Non-monetary contributions as a separate table
* [x] Add totals under the monetary tables?

### DIFF
--- a/api/subscribers/discourse.subscriber.js
+++ b/api/subscribers/discourse.subscriber.js
@@ -324,7 +324,7 @@ module.exports = {
       }
 
       if (expenses.length) {
-        content.push("#### Expenses");
+        content.push("#### Costs");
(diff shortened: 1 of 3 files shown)


Powersource:
so far

image

Powersource:
image

Powersource:
to get things into the non-monetary table you just put the value as 0 on a thing in the budget. we can maybe think of something nicer but this might also be enough, let's test it for a bit and see

Powersource:
image

PR #343, opened 2021-09-29
Upgrade react to 17

Powersource:
Fixes https://github.com/Edgeryders-Participio/multi-dreams/issues/342


Powersource:
As mentioned in the issue I've uninstalled the autoscroll dep used because it didn't support react 17 and it didn't seem to work where we used it in the app. The dev seems to have fixed the incompat now though so we could try re-adding it at some point if we want

https://github.com/brianmcallister/react-auto-scroll/issues/35

gustavlrsn:
The monster dialog is a bit weird without autoscroll so would be good to keep it

Powersource:
Yeah but that's the thing, it's not working at the moment :P I suggest we fix it in another PR but I can do it here if you want

gustavlrsn:
Ah, ok, gotya. Let's fix that in another PR then

PR #350, opened 2021-10-08
Use a wysiwyg editor for comments

Powersource:
Partially does https://github.com/Edgeryders-Participio/multi-dreams/issues/339

* [x] render comments with GFM. Fixed in https://github.com/Edgeryders-Participio/multi-dreams/pull/352
* [x] debounce onchange
* [x] make it less tall
* [x] the editor should also be used when editing an existing comment
* [x] the editor adds `<br>` when you add a bunch of newlines after eachother which don't render properly in the resulting comment. might be related to this bug https://github.com/nhn/tui.editor/issues/485
* [x] fix styling in preview to match the resulting comment
  * [x] fontsize even in un-previewed markdown should be bigger. let's skip this for now, seems convoluted
* [ ] hack in mobile support? or rip
* [ ] make sure the editor (in the different modes too) doesn't get too wide/fixed width so it pushes the bucket sidebar out. or too thin so the editor toolbar overflows
* [ ] can we have the pretty `color` outline like in prod?


aerugo:
Good job! It seems pretty smooth, but it would be nice to see how it works when going back in to edit, and how it looks when editing a comment between two other comments.

Powersource:
Yeah didn't record that, still To-do to enable

Jacob

Powersource:
the kind of thing that gives you confidence in the developers https://github.com/nhn/tui.editor/issues/1820

Powersource:
closing in favor of https://github.com/cobudget/cobudget/pull/360

PR #357, opened 2021-10-15
Only allow admins to create tags

Powersource:
Partially does https://github.com/Edgeryders-Participio/multi-dreams/issues/351

* [x] add mutation that admins can use to create tags
* [x] remove ability of addTag to create new tags
* [x] and update sidebar ui correspondingly
* [x] only display the tag sidebar if tags exist in the collection
* [x] move some files to ts


gustavlrsn:
There are linting errors

Powersource:
@gustavlrsn fixed

Powersource:
note to self, this allows me to see a very messy diff compared to the deleted file, idk if it's too messy to help though `g diff -w HEAD:api/resolvers/index.ts ui/server/graphql/resolvers/index.ts`

Powersource:
@gustavlrsn ready for review again

PR #360, opened 2021-10-28
Use remirror as a wysiwyg editor for comments

Powersource:
Notes for reviewers:
* Doesn't ssr https://github.com/remirror/remirror/issues/1349
* Doesn't support tables https://github.com/remirror/remirror/issues/1356
* Doesn't support task lists (checklists) https://github.com/remirror/remirror/issues/1357
* A bit bad performance when typing, hopefully not too bad https://github.com/remirror/remirror/issues/1366
* We need to get a new cloudinary preset (i think preferrably one that works for both comments as well as other future wysiwyg markdown fields) https://github.com/cobudget/cobudget/pull/360#discussion_r755441124

Partially does https://github.com/Edgeryders-Participio/multi-dreams/issues/339

* [x] fix toolbar on mobile (overflow scroll)
* [x] why does setting h3-6 (in the dropdown) submit the comment?? idk but fixed by removing the dropdown menu, the small headers are pretty pointless anyway
* [x] switch to the fancier table editor. ok both kinds of tables are a bit buggy so for now i've removed the button https://github.com/remirror/remirror/issues/1356
* [x] update markdown css class to handle loose lists better https://github.github.com/gfm/#loose
* [x] make editing text have same css as published comments. not literally the same, the html is different, but i fixed the list rendering at least
* [x] why do lists look weird? fixed by adding the normal md class?
* [x] add all the working options to the toolbar, such as lists and tables. removed the not working ones
* [x] make sure the sidebar doesn't get pushed out on medium size screens
* [x] fix border highlight color to match bucket
* [x] color on toolbar buttons
* [x] enable for editing comments as well
* [x] make cancel buttons work (incl clear)
* [x] fix autofocus prop
* [x] make sure it actually works for posting and editing comments
* [x] make sure all existing props are used
* [x] the editor when editing a comment shouldn't overflow into the sidebar
* [x] make short video to show how it looks
* [x] image embed support
* [x] switch to ts
* [x] images dragged and dropped from local should be stored on cdn
* [x] image upload button (so mobile can upload too)

inspo https://remirror.io/docs/showcase/markdown/


Powersource:
Some performance issues https://github.com/remirror/remirror/issues/1366 I think it should be fast enough but you'd have to try before merging @gustavlrsn

Powersource:
https://www.loom.com/share/55b0bdca7f6644a5abdf14095c353b34

ui/components/Wysiwyg.tsx
```diff
@@ -0,0 +1,341 @@
...
+      // timeout needed because a variable in the lib isn't init'ed yet
+      // see `uploads` here
+      // https://github.com/remirror/remirror/blob/3d62a9b937e48169fbe8c13871f882bfac74832f/packages/remirror__extension-image/src/image-extension.ts#L209-L214
+      setTimeout(() => setProgress(isUploading ? 0.4 : 1.0), 0)
+    );
+
+  return uploadImageFiles({
+    files,
+    setUploadingImages: setProgression,
+    cloudinaryPreset: "organization_logos", //TODO: get new preset
```
Powersource, line 126: @gustavlrsn

aerugo:
@gustavlrsn - looks like comments disappeared?
see https://cobudget.com/c/burberland/ckxbcosy5078208jux4vdlnat

Issue #469, opened 2021-11-30
Hide bucket review when a bucket is not published


gustavlrsn:
* [x] disallow reviewing unpublished buckets in api
* [x] hide monster when the bucket is unpublished

PR #479, opened 2021-12-07
Don't allow reviewing unpublished buckets

Powersource:
Fixes https://github.com/cobudget/cobudget/issues/469


Powersource:
ready for review again

Issue #456, opened 2021-11-30
Fix cache updates and pagination


gustavlrsn:
- [ ] collection member list
  - [x] update cache when approving members
  - [ ] fix pagination
- [x] when creating a coll
- [x] update cache when leaving a flag
- [x] clearing a flag
- [x] fund a bucket and then go to another page that lists how much funds you have left, the number has not been updated
- [x] create a bucket and it doesn't appear on the coll page

aerugo:
What's the status on this @Powersource ?

Powersource:
@aerugo i put it on ice because we were missing some member invite features but if that's in now i could pick it back up this week

PR #481, opened 2021-12-08
Fix cache updates and pagination

Powersource:
Fixes https://github.com/cobudget/cobudget/issues/456


Powersource:
we currently don't have the 'request to join' button so waiting with fixing that part

Issue #393, opened 2021-11-05
Implement email notifications for events according to specifications


aerugo:
As an organization admin, I want certain events to trigger notification emails. This issue also involves defining what those events are.

* [x] Invite_email (Join by invitation)
* [x] Confirm_account_email (Join by signing up yourself)
* [x] Welcome Email (after accepted invitation and after confirming email on self-sign-up) (only need to update text now)
* [x] ~Reset_password_email~ we don't have passwords (yet)
* [x] Notify_member_new_comment_in_bucket
* [x] Notify_member_new_comment_to_comment
* [x] Notify_member_that_they_received_allocation
* [x] Notify_funder_that_bucket_was_cancelled
* [x] Notify_member_new_funds_in_bucket (cocreator rather than member i guess) (do we actually want this every time? yes we'll do this for now)
* [x] Notify_member_new_bucket_in_group

tomomiq:
Links: 

Flows and triggers 
https://whimsical.com/emails-8up4tNoUDT2ubjNNm6xfXi

Template text 
https://docs.google.com/document/d/1O5ZAVGcye9gOI9RZrvDs0Vs9DgRRF23bHxnfVpflwUs/edit

Powersource:
@tomomiq in that first link, "prioritized" is "prioritized implementation" or?

tomomiq:
Yes. 
(And in Spain we agreed with Hugi that the Digests would be de-prioritized, so I changed that from red to grey just now.)

Powersource:
@tomomiq I've tried to combine the emails from the two links and made a list up top here, keeping only the prioritized ones. Can you check that it looks correct? Only one I'm unsure of is what the "New bucket online" in the graph means, I didn't include that one here now.

Powersource:
and what's our idea for "As a user, I want to be able to toggle which events lead to notification emails.", do we skip this initially? do i add a simple settings page with a checkbox for all of the above types?

aerugo:
A simple settings page is the right way to go. It should be reasonably easy
to implement a page with some toggles?

Powersource:
Well yes, but we also have like 3 weeks (if discounting holidays) until the milestone should be done 😅

aerugo:
Yeah, true. Leave the setting toggles for now, but make another issue for
them and push it to the next milestone.

DavidsBees:
[...]

PR #485, opened 2021-12-16
Email notifications

Powersource:
for https://github.com/cobudget/cobudget/issues/393

* [x] switch to nodemailer so we can use nodemailer-postmark-transport in prod but stmp in dev. that allows us to catch emails using mailhog in dev and inspect that they look correct
* [x] add basic sanitization
* [x] implement all email types (see issue)
* [x] test and verify that the postmark transport actually works


gustavlrsn:
LGTM 👏

Issue #444, opened 2021-11-22
Switch to AGPL3 + additional terms 


aerugo:
New license based on Pretix license: https://docs.pretix.eu/en/latest/license/faq.html

Powersource:
More related links

https://open.sentry.io/licensing/  
https://blog.sentry.io/2019/11/06/relicensing-sentry  
https://github.com/pretix/pretix/blob/master/LICENSE  
https://pretix.eu/about/en/blog/20210412-license/

Powersource:
when picking a new license i think it's good if we roughly specify what we want it to accomplish. @aerugo could you write a short list of things you want it to do?

Powersource:
also TODO: look at mongodb's license

aerugo:
@Powersource 
My thinking right now is that I would like to achieve this:

 - We own and control the name Cobudget, meaning that nobody else can sell services under that company name
 - We also name the software Cobudget and make sure that it is clearly identified as such in the frontend - with the name hard-coded and links pointing back to our landing page where we sell the SaaS. This means that if anyone self-hosts Cobudget, they will be advertising the cobudget.com service. 
 - If someone runs Cobudget unmodified, they can do whatever they want, including linking it to their own Stripe. However, they will then be advertising cobudget.com through doing so.
 - If someone runs a modified version of Cobudget, there are terms they need to follow. They must keep the link back to cobudget.com intact, and if they offer their service as a SaaS, they must also make the changes available. 

In my mind, this gives the community a lot of flexibility while making sure that they cannot simply rename the project "CowBudget" and start selling it as such. 

This is the model that Pretix has gone for: https://docs.pretix.eu/en/latest/license/faq.html

aerugo:
Note however that the above is basically just based on that I thought the Pretix license was pretty clever. I'm open to other strategies.

Powersource:
Haven't had time to read the pretix license thoroughly yet but I think for the above list of requirements we basically have all of them already with our trademark+agpl, except for the "They must keep the link back to cobudget.com intact". This is basically how mozilla does it with firefox, and they even have a looser license https://www.mozilla.org/en-US/foundation/trademarks/policy/

aerugo:
If you would want to look at the Pretix license additions to AGPL, that
would be quite useful.
But does our current license really make sure that someone can’t just
modify the code to rename the platform “CowBudget” and link to their own
SaaS sales funnel at cowbudget.com?

Powersource:
[...]

PR #492, opened 2021-12-21
Switch to new license based on the pretix license

Powersource:
fixes https://github.com/cobudget/cobudget/issues/444

* [ ] ask a lawyer?
* [ ] maybe get a :+1: from previous contributors, if we want this to apply to 'retroactive' code. but maybe/prob not worth it. if we don't, do we have to mention that old code has the old license or is git enough for that?
* [ ] should we mention trademark stuff anywhere? pretix only seems to do that in their faq

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -18,664 +18,664 @@ Public License are as follows:
 
 # Full text of the GNU Affero General Public License version 3
 
-                        GNU AFFERO GENERAL PUBLIC LICENSE
-                           Version 3, 19 November 2007
-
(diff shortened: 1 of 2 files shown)


aerugo:
I think it looks good, but no rush to merge. Thanks @Powersource! Let's talk about it over with the whole team and make sure this covers what we want it to cover.

Powersource:
discussed this in the monthly meeting and the meeting notes summed it up as "Consent on moving forward"

Issue #347, opened 2021-10-05
Add ability to @mention a fellow user in the Comments section


tomomiq:
(From the GT weekly call) 

As a logged in user, I can @mention someone In the comments section of the individual Bucket page so that I can engage them in conversation. 

Includes: 

- autocomplete of user names 
- text-style of usernames so it stands out a bit 
- In the future: an @mention would trigger a notification e-mail to that person 

We didn't make anything visual but hope that's sufficient description, let me know if there are other considerations for this feature!

Powersource:
the editor we're switching to has stuff for this https://remirror.io/docs/showcase/social/

Powersource:
@aerugo i don't remember what the latest discussion was on this around discourse, did we say we would just skip enabling this in orgs with discourse?

Powersource:
I just checked and mentions on discourse seem to be stored simply as e.g. `@person` in the markdown. I was thinking of making it stored ssb-style i.e. something like `[@person](https://cobudget.com/user/123asdfID)` since that would make rendering easier, it'd just be a basic link. But then discourse doesn't seem to be counting that as a mention in their book. Do you have a strong preference for the former? Should we or discourse be the ones to handle notifications when the org has discourse?

aerugo:
For groups using Discourse, most people will be commenting through Discourse and only rarely through the Cobudget interface itself. When people write comments on Discourse they will then not render mentions on Cobudget even when the username is the same. 

I can live with that if we have a strong preference for the ssb-style. I think Discourse communities will be rare enough that it's not a big deal.

aerugo:
Actually, we could solve that by converting @hugi in a Discourse comment to [@hugi](https://forum.blivande.com/u/hugi/summary) in the client when we render a Discourse comment. That is probably the most elegant solution. That way the mention has some purpose when rendered on Cobudget - it links to a profile on Discourse.

PR #525, opened 2022-01-24
Mentions in comments

Powersource:
For https://github.com/cobudget/cobudget/issues/347

* [x] make basic user page to be able to link to
* [x] make sure it doesn't consume all `[link](https://example.com)` style links (probably need to narrow down the selector)
* [x] make sure it's only available in comments
* [x] should be able to mention actual users (search through round members but link to user)
* [x] don't error the user page if invalid id
* [x] remove the imported extension in case we don't end up using it. including the babel config and tsconfig things
* [x] style mentions a bit differently in Markdown component rendering
* [x] send an email to the person being mentioned (if they were already gonna get an email, don't send that other one)
* [x] render comments in emails
* [x] reenable mailhog testing
* [x] the popup sometimes ends up being covered by the save button. do we wanna hack around this while we wait for a fix? https://github.com/remirror/remirror/issues/1511


Powersource:
Made an issue in remirror https://github.com/remirror/remirror/issues/1479

gustavlrsn (commented):
LGTM

Issue #534, opened 2022-01-31
[FEATURE] Ability to "re-send" invite emails to members of a round and group


francesca-sp:
**Describe the feature**
Right now, it is only possible to send out an invitation to people to join a Cobudget round or group once. This is problematic, because it is common that such an invite email might get lost, it ends up in a spam folder or is deleted by mistake. Therefore it is an important feature for the group admin to be able to resend an invite. 
![image](https://user-images.githubusercontent.com/3641828/151818580-15048cec-a657-4fc5-bd50-7535d8bb9dd3.png)

It could look similar to in legacy cobudget. When you click on the three dots on the right of a member, you add an additional action next to "make admin" called "invite again" or "resend invite". 

![image](https://user-images.githubusercontent.com/3641828/151818305-cc8c5573-7ac4-42ab-b6bb-0b000b7d2f4f.png)

aerugo:
This is a duplicate of #443 - I will move it to Member management milestone and replace that issue with this one.

gustavlrsn:
The workaround of simply inviting them again is working for Rounds but not Groups, which is a bug tracked here #535

alinauroz:
What is the correct way to detect a pending invite?

If an invite is sent to a user who has not joined Cobudget, it means the invite is pending. Right?

aerugo:
An invite should be pending until the user clicks the link, and this action should change the users membership in the round from pending.

Users may still have joined Cobudget and be pending on a specific round, because the invite is for them to join that round, not Cobudget as a whole.

aerugo:
Note that this is different than “verified” which a user only needs to do once to verify their email once they register on cobudget. So there are two states: Verified (global) and pending (local).

alinauroz:
Right now, when an invitation is sent to a user who is already registered on cobudget, no approval is required. 

For example ali@ex.com is a member of cobudegt. When an invitation is sent to ali@ex.com, this user is immediately added to the collection.

alinauroz:
Approvals are only pending when a user is has not joined cobudget.

aerugo:
Yes, this is true. A better system for handling pending users is on the TODO. For now, I would add a simple “pending” to the member table and flip that with a link in the invite email.

aerugo:
I have discussed with @alinauroz and this issue is actually a little larger in scope than we thought.
We need to add an invitations table to store unique invite links to the collection. He will go ahead and do that too.

alinauroz:
Final Solution:

[...]

PR #537, opened 2022-02-05
Feat/invite again

alinauroz:
This MR fixes https://github.com/cobudget/cobudget/issues/534

### DIFF
--- a/ui/graphql/client.ts
+++ b/ui/graphql/client.ts
@@ -81,6 +81,16 @@ export const client = (
                 );
               }
             },
+            acceptInvitation(result: any, args, cache) {
+              if (result?.acceptInvitation?.hasJoined) {
+                cache
+                  .inspectFields("Query")
+                  .filter((field) => field.fieldName === "membersPage")
+                  .forEach((field) => {
+                    cache.invalidate("Query", "membersPage", field.arguments);
+                  });
+                }
+            },
             joinOrg(result: any, args, cache) {
               if (result.joinOrg) {
                 cache.updateQuery(
--- a/ui/graphql/client.ts
+++ b/ui/graphql/client.ts
@@ -338,12 +338,24 @@ export const client = (
                     variables: { collectionId, offset: 0, limit: 1000 },
                   },
                   (data: any) => {
+
+                    const existingEmails = data.approvedMembersPage?.approvedMembers
+                                              ?.map(member => member.email) || [];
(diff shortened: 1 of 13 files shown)


ui/graphql/client.ts
```diff
@@ -328,12 +338,24 @@ export const client = (
...
+                                              ?.map(member => member.email) || [];
+                    const newInvitedMembers = result.inviteCollectionMembers
+                                              ?.filter(member => (
+                                                existingEmails.indexOf(member.email) === -1
+                                              ));
+
+                    if (newInvitedMembers.length === 0) {
+                      return;
+                    }
+
```
gustavlrsn, line 352: What is this code trying to do? Having a hard time to follow, either we add a comment, or we could also invalidate the cache instead to simplify this.

ui/server/prisma/migrations/20220202110646_collection_member_add_has_joined_col/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "CollectionMember" ADD COLUMN     "hasJoined" BOOLEAN NOT NULL DEFAULT false;
```
gustavlrsn, line 2: This is causing all existing members to have `hasJoined` default to `false`. The subsequent migration won't change that since all fields will already be set from this migration. You can delete both migration files and generate one again, that have it set the default to `true`

aerugo:
> @aerugo do we want the `invite again` feature also for Groups a.k.a Organizations?

Yes, but it does not have to be part of this PR, we could add that later.

ui/graphql/client.ts
alinauroz: https://user-images.githubusercontent.com/38759611/152821570-ebda3b94-ed98-4d80-9a53-e96a95abcaf5.mp4

There was an issue. When you invite the members, the invited members array was concatenated with the existing members without checking whether the new invited members are in the list of existing members or not.

In these lines, I check for existing members and don't push new invited member if it is already there.

ui/server/prisma/migrations/20220202110646_collection_member_add_has_joined_col/migration.sql
alinauroz: Alright, I'll do this

Issue #369, opened 2021-11-05
Search for members on paginated member page


aerugo:

aerugo:
related https://github.com/Edgeryders-Participio/multi-dreams/issues/307 /j

Powersource:
fwiw there's some member searching added to the api in https://github.com/cobudget/cobudget/pull/525 that might be reusable/usable for inspiration

aerugo:
Ping @SeunghunSunmoonLee

SeunghunSunmoonLee:
Just followed README and I get error. 
`➜  ui git:(main) ✗ yarn dev
yarn run v1.22.4
$ yarn run generate
$ yarn run generate:prisma
$ prisma generate
Environment variables loaded from .env
Prisma schema loaded from server/prisma/schema.prisma

✔ Generated Prisma Client (3.4.1) to ./node_modules/@prisma/client in 344ms
You can now start using Prisma Client in your code. Reference: https://pris.ly/d/client
```
import { PrismaClient } from '@prisma/client'
const prisma = new PrismaClient()
```
$ next dev
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
info  - Loaded env from /Users/seunghunlee/Software/cobudget/ui/.env.local
info  - Loaded env from /Users/seunghunlee/Software/cobudget/ui/.env.development
info  - Loaded env from /Users/seunghunlee/Software/cobudget/ui/.env
info  - Disabled SWC as replacement for Babel because of custom Babel configuration ".babelrc" https://nextjs.org/docs/messages/swc-disabled
event - compiled client and server successfully in 2.9s (2113 modules)
wait  - compiling / (client and server)...
info  - Using external babel configuration from /Users/seunghunlee/Software/cobudget/ui/.babelrc
event - compiled client and server successfully in 1483 ms (2116 modules)
wait  - compiling /api...
event - compiled client and server successfully in 431 ms (2160 modules)
wait  - compiling /_error (client and server)...
error - ReferenceError: Cannot access '__WEBPACK_DEFAULT_EXPORT__' before initialization
event - compiled client and server successfully in 292 ms (2161 modules)`

Powersource:
[...]

PR #545, opened 2022-02-10
Add search for members on paginated member page (#369)

SeunghunSunmoonLee:


README.md
```diff
@@ -14,6 +14,8 @@ Next.js, Urql, Apollo, Prisma, PostgreSQL
 - Install dependencies in `/ui`
   - `yarn`
 - In /ui, copy `.env.local.default` to `.env.local`
+- In /ui, copy and paste DATABASE_URL string from `.env.development` to `.env.local`
```
Powersource, line 17: I don't think this should be needed? Both files are loaded.

But ah on checking `.env.local.default` we have `DATABASE_URL=""` in it which clears the variable. I think a better solution is to just remove that line.

README.md
```diff
@@ -14,6 +14,8 @@ Next.js, Urql, Apollo, Prisma, PostgreSQL
...
+- put MAGIC_LINK_SECRET="fake" in `.env.local`
```
Powersource, line 18: It's easier if this is added to `.env.local.default` instead of the readme.

ui/components/EventMembers/index.tsx
```diff
@@ -114,11 +118,30 @@ const EventMembers = ({ collection, currentUser }) => {
...
+
+  const items = useMemo(() => {
+    if (loading || !approvedMembers) {
+      return [];
+    }
+    return approvedMembers
+  }, [approvedMembers, loading]);
+
+  useEffect(() => {
+    // if (tooShortSearch) return;
```
Powersource, line 140: don't we want this limitation?

ui/components/EventMembers/index.tsx
```diff
@@ -114,11 +118,30 @@ const EventMembers = ({ collection, currentUser }) => {
       fetching: loading,
       error,
     },
+    searchApprovedMembers
   ] = useQuery({
     query: COLLECTION_MEMBERS_QUERY,
-    variables: { collectionId: collection.id, offset: 0, limit: 1000 },
+    variables: { collectionId: collection.id, usernameStartsWith: searchString, offset: 0, limit: 1000 },
```
Powersource, line 124: @aerugo is it enough for now that we search by username or do we want e.g. name and/or email as well?

ui/components/SearchBar/index.js
```diff
@@ -0,0 +1,38 @@
+import React, { useEffect, useRef } from "react";
+import { SearchIcon, CloseIcon } from "../Icons";
+
```
Powersource, line 3: We want to slowly move to ts for all files so better to have that by default when creating a new file. If you run into any annoying type issues it's fine to use `any` though

ui/components/SearchBar/index.js
```diff
@@ -0,0 +1,38 @@
...
+export default ({ placeholder, collection, clearInput, value, onChange }) => {
+  const inputRef = useRef();
+  useEffect(() => {
+    inputRef.current.focus();
+  }, []);
```
Powersource, line 8: i think we should use `autoFocus` on the element instead

Powersource (changes requested):
Saw too late that it was still WIP so feel free to ignore the comments you were gonna adress anyway :grin:

SeunghunSunmoonLee:
All suggestions are applied [d507619 ](https://github.com/cobudget/cobudget/commit/d5076190a4436f9063aad3cc9d71d9c8d403a14f)[now](https://github.com/cobudget/cobudget/pull/545#event-6051365199)

ui/components/EventMembers/index.tsx
aerugo: Lets do username for now, then add name and email later.

ui/components/EventMembers/index.tsx
aerugo: Yeah, I think one character makes sense when filtering a list where it's fine to have many results, but two characters are better for mentions - where you definitely never want to fill the autocomplete selector with all usernames starting with "a".

aerugo:
@Powersource - you think we're ready to merge?

Powersource:
merged this now, but with a few more commits. had to make the changes outside of this pr since i didn't have push access to this branch. feels like we should figure out a smoother collab setup for the next pr.

https://github.com/cobudget/cobudget/commit/80d6a47b6a1cad3ba4b7171b458a609bde355fe4
https://github.com/cobudget/cobudget/commit/628c27c677d1dd46ac64b56317bcb795657073a7
https://github.com/cobudget/cobudget/commit/de5d80f490eb9d0b90a4bbeb97746c140a0195c1

Issue #543, opened 2022-02-08
[BUG] Questions default to short text when edited


skoteskote:
**Describe the bug**
Settings - Questions - Question modal defaults back to Short text after any edit of question.

**To Reproduce**
1. Go to Settings - Questions for a round.
2. Add a question, set it to Long text.
3. Save.
4. Edit the question and change eg the title. 
5. The question automatically change from long text to short text.

**Expected behavior**
The question should not change the desired text length without explicit user interaction.

**Browser and device (please complete the following information):**
 - Browser Brave, Safari
 - Device and/or OS: Mac

skoteskote:
@aerugo Your choice where to put it, but it's very annoying and I guess an easy fix.

aerugo:
Seems like a pretty critical bug, so I put it in our current milestone.

tomomiq:
Thank you @alinauroz !

PR #560, opened 2022-02-16
fix: init typeInputValue

alinauroz:


alinauroz:
Fixes https://github.com/cobudget/cobudget/issues/543

Issue #441, opened 2021-11-22
Integrate Helpscout 


aerugo:
<img width="1295" alt="Screenshot 2022-02-09 at 23 23 29" src="https://user-images.githubusercontent.com/7785081/153307360-ba9e9f92-898f-4175-89ff-cfd91524be7d.png">

_Screenshot from the old version of Cobudget_

We want to integrate HelpScout. Let the support widget be visible on all pages, and allow setting the HelpScout key through .env. If there is no HelpScout key, disable HelpScout.

Powersource:
@skoteskote might have to look at this since I think these integrations tend to add tracking.

also fwiw I often don't see these helpers at all since they're often blocked by extensions that block trackers.

francesca-sp:
@Powersource we just talked about the soft launch plans and think it would be important to integrate this by March 1st. It's really easy to do, I remember it was when we did it with legacy Cobudget. Here is the code snippet that needs to be integrated. 

**Paste this code before the </body> tag:**

`<script type="text/javascript">!function(e,t,n){function a(){var e=t.getElementsByTagName("script")[0],n=t.createElement("script");n.type="text/javascript",n.async=!0,n.src="https://beacon-v2.helpscout.net",e.parentNode.insertBefore(n,e)}if(e.Beacon=n=function(t,n,a){e.Beacon.readyQueue.push({method:t,options:n,data:a})},n.readyQueue=[],"complete"===t.readyState)return a();e.attachEvent?e.attachEvent("onload",a):e.addEventListener("load",a,!1)}(window,document,window.Beacon||function(){});</script>
<script type="text/javascript">window.Beacon('init', 'e1d18cac-eb2f-43c0-a109-6b185575b27e')</script>`

Let me know if you need anything else, or

francesca-sp:
this looks good to me! ping @tomomiq

eden-gt:
looks good!

francesca-sp:
Looks good! Closing the issue.

PR #561, opened 2022-02-16
helpscout integration

alinauroz:
Fixes https://github.com/cobudget/cobudget/issues/441

### DIFF
--- a/ui/.env
+++ b/ui/.env
@@ -1,3 +1,4 @@
 KEYCLOAK_CLIENT_ID="dreams"
 KEYCLOAK_REALM=plato
 DATABASE_URL='postgres://postgres:postgres@localhost:35432/cobudget'
+HELPSCOUT_KEY=e1d18cac-eb2f-43c0-a109-6b185575b27e
--- a/ui/components/Layout.js
+++ b/ui/components/Layout.js
@@ -1,6 +1,7 @@
 import React from "react";
 import Head from "next/head";
 import Header from "./Header";
+import "../lib/beacon";
 
 const LinkOut = ({ href, children }) => {
   return (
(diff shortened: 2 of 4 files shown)


ui/.env
```diff
@@ -1,3 +1,4 @@
 KEYCLOAK_CLIENT_ID="dreams"
 KEYCLOAK_REALM=plato
-DATABASE_URL='postgres://postgres:postgres@localhost:35432/cobudget'
\ No newline at end of file
+DATABASE_URL='postgres://postgres:postgres@localhost:35432/cobudget'
+HELPSCOUT_KEY=e1d18cac-eb2f-43c0-a109-6b185575b27e
```
Powersource, line 4: I think it's better to remove this from here and supply it in vercel instead. i don't think we want other people deploying the app to send their users to our support.

I just added it to vercel now so it should be safe to remove from here.

also check that the app still work if the variable isn't supplied

Powersource (changes requested):

Powersource:
image

also i'm not entirely a fan of how this looks on mobile (or these kinds of bubbles in general) but this is probably the wrong place to take that discussion :P At least it's much smaller than the old review button that we used to have.

But I think we at least should add some padding below the normal footer so that you can see the Terms link again.

alinauroz:
I'll add extra padding to the footer design.

ui/.env
alinauroz: If HELPSCOUT_KEY is missing, then the Helpscout Beacon is not shown. So if i remove it locally, Beacon will not be shown on the page. The app will still be working.

Powersource: yep that's good but i think we want to remove the line from this file

Issue #366, opened 2021-11-05
Display user available balance for round


aerugo:

tomomiq:
Assigning to myself to make a sketch

tomomiq:
**Round page:**

- It shouldn't be on all Round pages, just the ones that have a bucket that can be funded now/soon. But that could be de-prioritized re: difficulty in assessing status? 
- Using the currency symbol "€" instead of the typed out word "euro", to make it more scannable and would like this to be rolled out across the app eventually.  

<img width="935" alt="Screenshot 2022-02-01 at 16 17 54" src="https://user-images.githubusercontent.com/1287179/151997905-1455b551-585d-48f2-970e-3bcb96651ed8.png">

**Bucket page**

- This page wasn't part of the original scope of this issue but Fran pointed out that it's more important here so here's a sketch! 

<img width="431" alt="Screenshot 2022-02-01 at 16 27 53" src="https://user-images.githubusercontent.com/1287179/151997943-186cc6ec-6a20-461b-b079-d264fa45125c.png">

tomomiq:
Hi @alinauroz cc @Powersource 

Seems there is some confusion of how these numbers are calculated so we would like to re-write the copy like this: 

<img width="824" alt="Screenshot 2022-02-22 at 15 59 57" src="https://user-images.githubusercontent.com/1287179/155159327-77e808d1-a980-41da-a47b-261cd03c026e.png">

It should be 

* the amount that is available to that logged-in user to fund something 
* the total of what is available (i.e. not assigned to any particular bucket yet) for the round, across all users 

Also, is there a border style that can be re-used here, to differentiate that this element is not part of the bucket?

tomomiq:
Switching to Bug status, and priority for  fixing. 
We already had a user report that the calculation is wrong.

Powersource:
@tomomiq i think it's correct now, please verify

Powersource:
\*in a few min when it's deployed

tomomiq:
Great!

PR #566, opened 2022-02-18
Feat/show balance

alinauroz:
fixes https://github.com/cobudget/cobudget/issues/366


Powersource (commented):
looks great!

i was a bit confused by the "123€ in round" part but that was a part of the design, so let's try it in prod and see how it is to use it

Issue #515, opened 2022-01-17
Crash when visiting bucket in hidden round


aerugo:
Steps to reproduce:

- Set a round to "hidden"
- Go to a bucket through direct link

Result:
- Vercel crash screen 

Suggested solution:
First step: Solve crash and make sure it directs to somewhere.
Later:
All attempts to visit a hidden round or the content from a hidden round should go to a nice friendly message screen. Perhaps @tomomiq could help design something nice?

Powersource:
also fixed so hidden rounds don't get shown to everyone on the group page

Powersource:
also when testing this, go to an actually hidden bucket (and not just change the url) because you might run into https://github.com/cobudget/cobudget/issues/592

tomomiq:
hi @Powersource 

It looks like the hidden Bucket is taken care of (second image) but not a hidden Round (first image). 
Can you confirm if you worked on both? 

Also, if you could add more padding above the error message, it would look a little nicer! :) 

<img width="1354" alt="Screenshot 2022-03-04 at 11 42 03" src="https://user-images.githubusercontent.com/1287179/156748827-8b4f42f2-99e1-4498-ac10-ca06781c2022.png">
<img width="1354" alt="Screenshot 2022-03-04 at 11 41 59" src="https://user-images.githubusercontent.com/1287179/156748839-d618aed0-2ffb-4a5c-b9d7-35b711cdf04a.png">

Powersource:
No I didn't work on the round page, wasn't mentioned. I'll fix that now.

About the padding, I think that's better left to https://github.com/cobudget/cobudget/issues/583

I think there's some drift here from the original issue, and I think it's interesting to think about the prioritization of these kinds of fixes versus things we've prioritized in the backlog.

Powersource:
fixed in https://github.com/cobudget/cobudget/commit/3a2f07d1d627825c0f5c222a1b0a02a649649f36

francesca-sp:
@Powersource I agree that adding padding is out of scope of the original issue. In terms of also fixing this for the rounds scenario and not just buckets, to me it seems obvious that if you prioritize one, you also prioritize the other. In any case, thanks for fixing it.

francesca-sp:
@Powersource working correctly!

PR #569, opened 2022-02-22
Just show blank page when visiting bucket in hidden round

Powersource:
for https://github.com/cobudget/cobudget/issues/515

* [x] don't list hidden rounds to strangers on group page


gustavlrsn:
LGTM

Issue #586, opened 2022-03-02
[BUG] Can't delete member


alanna:
I added my alt email as a test account and now I want to remove it. 

1. I go to members > delete
2. A popup appears indicating the email is 'undefined'
3. I click OK, but the member is not deleted

<img width="713" alt="Screen Shot 2022-03-02 at 2 25 26 PM" src="https://user-images.githubusercontent.com/5498878/156277380-40a5f944-6711-4b96-ae13-27ffc31b8637.png">

<img width="442" alt="Screen Shot 2022-03-02 at 2 26 52 PM" src="https://user-images.githubusercontent.com/5498878/156277372-cdde6b42-1640-4f10-8ffd-58c13df5066c.png">

alanna:
There also doesn't seem to be a way for me as a user to leave the group, when logged into the test account.

Powersource:
thanks for the report, taking a look

Powersource:
At first when testing, I was able to delete a coll member, but when I tried it again with a user that had money allocated to them it broke with

> "Foreign key constraint failed on the field: `Allocation_collectionMemberId_fkey (index)`"

that makes some ux sense but I also get the same error after I've set the user's balance back to 0 (I think this is Alanna's case)

gustavlrsn:
Yeah, if balance is zero it would be good to be able to remove a member. 

Either we do a soft delete or connect allocations/contributions to userId instead of collectionMemberId.

But I think it makes most sense to connect allocations/contributions to userId instead of collectionMemberId as it creates less nested data dependencies. Same thing for comments and maybe other stuff. Working on a fix.

francesca-sp:
Just as an FYI that we received another bug report on this same issue from a user yesterday at 8 pm @Powersource

aerugo:
It still seems to be a problem. 
Two issues, not sure if related:
 - email is shown in message as undefined 
 - user is not deleted - regardless of if the balance is 0.

<img width="1840" alt="Screenshot 2022-03-23 at 16 59 02" src="https://user-images.githubusercontent.com/7785081/159742726-f2d4bc66-55e0-4d09-bdd6-7a43d226cf3a.png">

PR #588, opened 2022-03-02
Can't delete round members

Powersource:
for https://github.com/cobudget/cobudget/issues/586

* [x] don't show isRemoved members in the "to approve" list
* [x] only allow removing people if they have 0 balance
  * [x] api
  * [x] ui
* [x] go through the different member-adding paths and make sure they work for a user who used to be a member
* [x] what about members with "invitation pending"


ui/components/Header/index.tsx
```diff
@@ -160,36 +163,40 @@ const Header = ({ collection, currentUser, currentOrg, openModal, router }) => {
...
-                              toast.error(error.message);
-                            } else {
-                              toast.success(
-                                collection.registrationPolicy ===
-                                  "REQUEST_TO_JOIN"
-                                  ? "Request sent!"
-                                  : "You joined this collection!"
-                              );
+                {
+                  // you can ask to be a member if you've either never been a member, or been a member but been removed
```
Powersource, line 167: doublecheck the logic here before merging

Powersource: guarded a bit more now

Powersource:
@gustavlrsn I've tried to be careful with the logic and have tested all the user flows I could think of but if you could ctrl-f through the diff for `isApproved` `hasJoined` and `isRemoved` and see that the logic makes sense then that'd be nice. It's also possible I've forgotten some case where we use these variables and we'd need to update the logic, but since I've sort of "hid" isRemoved behind isApproved (see comment in the schema) I think we should be fine.

Powersource:
also this pr inspired https://github.com/cobudget/cobudget/issues/627

gustavlrsn:
LGTM! This seems like the right approach 👍

gustavlrsn:
Some merge conflicts

Issue #582, opened 2022-03-01
[BUG] If I delete myself from a round, application crashes (on members + transaction pages)


eden-gt:
**Describe the bug**
If I delete myself from a round, application crashes (on members + transaction pages)

**To Reproduce**

1. Go to a members tab of a round 
2. Click on yourself
3. Press Delete
4. Confirm delete with pop up 
5. App crashes

**Expected behavior**
A blank page telling me:
"Oops, something went wrong! Go back to the homepage".

**Screenshots**
<img width="940" alt="Screenshot 2022-03-01 at 14 56 35" src="https://user-images.githubusercontent.com/94849002/156192972-86325e06-328e-43d8-979d-90284c209db4.png">

<img width="932" alt="Screenshot 2022-03-01 at 14 54 50" src="https://user-images.githubusercontent.com/94849002/156192997-ddb3a041-a350-451c-9563-32cf34dbf8d1.png">

**Browser and device (please complete the following information):**

- Browser: chrome
- Device and/or OS:  Mac & PC

tomomiq:
Confirmed I could successfully delete myself.

PR #606, opened 2022-03-14
fix/delete-self

alinauroz:
For https://github.com/cobudget/cobudget/issues/582

Issue was default values are used in case of undefined not in case of null.

Also fixed the confirmation message

`Are you sure you would like to delete membership from user with email undefined`

to

`Are you sure you would like to delete membership from user with email <email>`


ui/components/EventMembers/index.tsx
```diff
@@ -135,6 +126,14 @@ const EventMembers = ({ collection, currentUser }) => {
     pause: true,
   });
 
+  useEffect(() => {
```
Powersource, line 129: Can we simply have simple variables instead of extra react state for this? E.g. `const approvedMembers = data?.approvedMembersPage?.approvedMembers || [];`

alinauroz: Yes, we can do it. This approach is much simpler. 

This way i think there is a small advantage. For example, is moreExists is false, we can search(filter) from local data instead of calling api.(right now api is called).

We may want to implement pagination in future. So if we add pagination, `const approvedMembers = data?.approvedMembersPage?.approvedMembers || [];` will only return limited numbers of result.

For example, 10 results are loaded on page load. When the window is scrolled down, 10 more results are fetched. In this case, `const approvedMembers = data?.approvedMembersPage?.approvedMembers || [];` will override the previously loaded data.

By using a separate state, we can append the newly added loaded data to the state.

gustavlrsn: I think @Powersource's suggestion is the right one, it seems simpler to me.

Also, currently the strategy to do pagination is to add another Page component, and not append individual items to an array in state, see the buckets list for example.

And also, I think searching should be done by the API as it is now, it's more likely that we want to search the entire database records and not only filter the items currently displayed on the page.

ui/components/EventMembers/index.tsx
```diff
@@ -152,14 +144,26 @@ const EventMembers = ({ collection, currentUser }) => {
...
+      return _member.id === deleteMemberResponse?.deleteMember.id
+    });
+    
+    if (!deleteMemberLoading && (currentUser.id === member?.user?.id)) {
+      Router.reload();
+    }
+
+    console.log(deleteMemberResponse, currentUser);
+  }, [deleteMemberLoading]);
+
```
gustavlrsn, line 166: If the purpose here is to remove the deleted member from the list, this should be done by updating the urql cache.

aerugo:
@alinauroz -are the changes requested by @gustavlrsn done?

alinauroz:
> @alinauroz -are the changes requested by @gustavlrsn done?

Yes, done

ui/components/EventMembers/index.tsx
Powersource: added logic for clearing the cache now

Issue #559, opened 2022-02-15
[MINOR] Copy changes for terminology clean up


eden-gt:
Capturing here all the places where terms need changing to reflect terminology:

**Collection to be replaced with Round**
Exact places:
- Round Settings > General:  
<img width="888" alt="Screenshot 2022-02-15 at 14 32 31" src="https://user-images.githubusercontent.com/94849002/154083387-8e412774-e726-4030-9ddd-0d894c4d6c25.png">
- Invite members' pop up (which also might need to have the "beta" text removed?)
<img width="622" alt="Screenshot 2022-02-15 at 14 52 37" src="https://user-images.githubusercontent.com/94849002/154087396-9093d5ef-b312-40d4-97bc-4c045507b008.png">

- In Group main page, the button should say "New Round" (not "new collection")
<img width="1038" alt="Screenshot 2022-02-15 at 15 01 38" src="https://user-images.githubusercontent.com/94849002/154088805-4504a309-1575-4316-8602-4645d4f7e7f6.png">

**"Granting" should always be replaced with "Funding"**
Exact places where this is:
- Round "About" Page -
<img width="837" alt="Screenshot 2022-02-15 at 14 27 57" src="https://user-images.githubusercontent.com/94849002/154082158-30402467-3bf3-4d3d-970b-50421a792a90.png">

- Granting Settings Page (within Round Settings)
<img width="742" alt="Screenshot 2022-02-15 at 14 33 33" src="https://user-images.githubusercontent.com/94849002/154088346-2fb8153b-73cb-43b4-a5f4-56bdaab387a4.png">

- Round Settings Page, in the left hand side menu.
<img width="577" alt="Screenshot 2022-02-15 at 14 19 01" src="https://user-images.githubusercontent.com/94849002/154082683-7af65a35-1f06-4d2d-83b4-4acf14f4c148.png">

**Questions to be renamed "Bucket Form"** 
See screenshot above

**Members should be renamed "Participants" in round (we call them members only at Group level)**
<img width="1036" alt="Screenshot 2022-02-15 at 15 03 41" src="https://user-images.githubusercontent.com/94849002/154089041-a1252344-07ae-4460-bbe2-63a1677d093b.png">

tomomiq:
Please also review all of the email notification copy, to replace all mentions of collections with rounds.

aerugo:
Covered by #597

tomomiq:
Confirming that "collection" has been updated to "round". 

[...]

PR #632, opened 2022-03-19
Copy changes

alinauroz:
fixes https://github.com/cobudget/cobudget/issues/559


ui/components/SubMenu.tsx
```diff
@@ -19,7 +19,7 @@ const groupItems = ({ currentUser, groupSlug }) => {
   return [
     { label: "Overview", href: `/${groupSlug}` },
     // { label: "Realities", href: "/realities" },
-    { label: "Members", href: `/${groupSlug}/members`, admin: true },
+    { label: "Participants", href: `/${groupSlug}/participants`, admin: true },
```
Powersource, line 22: i think on the group level they should still be called Members, it's only in the round that they're called Participants

Powersource (changes requested):

Issue #600, opened 2022-03-10
[BUG] Application Error on members & transactions page that I don't have access to


francesca-sp:
**Describe the bug**
Error when going to certain round pages if you are a group admin who is not a member of the round. 

**To Reproduce**

1. Go to a public 'Round' you are not a member of, in a group where you are an admin
2. Click on ''Members" and "Transactions" and Settings
3. See application error shown below for members and transactions page, and on the settings pages you see the full page and can edit.

**Expected behavior**
When entering a round I am not a member of, I expect to not see tabs I don't have access to, in this case the Transactions and Settings tab. When going on the members tab, I should see the message telling me I cannot see it, because I have not joined ("You need to be a member of this round to see the member list). It needs to look the same as what is visible for users who are not logged in (in a public group). 

**Screenshots**
![image](https://user-images.githubusercontent.com/3641828/157703464-855c542d-0aad-4bae-83c9-f5ea8d3e572d.png)

Powersource:
on staging in a few min

eden-gt:
This is fixed!

PR #633, opened 2022-03-20
Fix/group admin auth

alinauroz:
For https://github.com/cobudget/cobudget/issues/600.

This MR fixes the issue and implements the expected behavior. This MR disables some group admin privileges who are not a member of a round.


ui/server/graphql/resolvers/index.ts
```diff
@@ -56,11 +56,14 @@ const isMemberOfGroup = async (parent, { groupId }, { user }) => {
...
-    bucketId,
-  });
+  let roundMember;
+  try {
+    roundMember = await getRoundMember({
+      userId: user.id,
+      roundId,
+      bucketId,
+    });
+  } catch (err){}
```
Powersource, line 66: Why do we need to catch and discard this?

Powersource: preferrably refactor to not need to ditch the error but otherwise at least add a comment for why we do it

alinauroz: The reason for catching the error was because we may want to continue execution even if there is error.

But the error is not ignored. If there is any error, the value of roundMember will be undefined which will be handled in next lines of code.

alinauroz:
Also Fixes https://github.com/cobudget/cobudget/issues/636

ui/server/graphql/resolvers/index.ts
Powersource: @alinauroz can you take care of this one?

alinauroz: > @alinauroz can you take care of this one?

yes

alinauroz: done

Issue #611, opened 2022-03-15
[FEATURE] Log in with Facebook and Google


skoteskote:
**Describe the feature**
A user should be able to log in with a Facebook or Google account. We briefly had this in the old version but it got removed I think? 

This is especially useful for use cases like eg Malmö Tillsammans where they have very large groups of users who have very low technical knowledge. This might surprise you, but for some people "just check your mail!" isn't as straightforward as it is for us. The ability to log in with one's Facebook account lowers the entry threshold immensely. 

* [ ] facebook https://github.com/cobudget/cobudget/pull/642
* [ ] google https://github.com/cobudget/cobudget/pull/658

Powersource:
"This redirect failed because the redirect URI is not whitelisted in the app’s Client OAuth Settings. Make sure Client and Web OAuth Login are on and add all your app domains as Valid OAuth Redirect URIs."

Powersource:
"The domain of this URL isn't included in the app's domains. To be able to load this URL, add all domains and subdomains of your app to the App Domains field in your app settings."

Powersource:
note, when testing this in staging eventually:
* test the flow where a user denies the `email` permission
* test signing up using a facebook account with an email address that doesn't yet have a normal cobudget account

Powersource:
facebook login can be tested now. see comment above for testing tips. if testing passes then don't close the issue (since we still have google left) but instead check the box in the issue description

aerugo:
I get this error when I try to sign up with Facebook. This Facebook account is associated with an email that is not registered on Cobudget staging or production.

<img width="1287" alt="Screenshot 2022-03-31 at 14 26 23" src="https://user-images.githubusercontent.com/7785081/161054720-a381163f-9da5-4074-845b-b2412f706a10.png">

Powersource:
@aerugo A fix should be deployed on staging now

Powersource:
google sign in is also on staging now

tomomiq:
Confirming that logging in with Facebook and Google accounts, when I have a Cobudget account with the same e-mail, works. 

@Powersource can this say Cobudget instead of Plato Project..? 

<img width="1365" alt="Screenshot 2022-04-04 at 22 22 17" src="https://user-images.githubusercontent.com/1287179/161625751-a248481c-9743-4c90-b9fc-0da08a4219f7.png">

tomomiq:
[...]

PR #642, opened 2022-03-27
Log in with facebook

Powersource:
For https://github.com/cobudget/cobudget/issues/611

* [x] make sure to send welcome email still
* [x] make sure app still runs when you don't have the env vars
* [x] add env vars to staging and prod vercel
* [x] add redirect urls for staging and prod to fb settings
* [x] make ui prettier
* [x] grep for TODO
* [x] take a look at how the Remember me logic interacts with this
* [x] look at replacing Remember me with instead pinging fb's servers now and then to see if the user is still logged in


Powersource:
"Authentication Expiration

When your app uses Facebook Login to authenticate someone, it receives a User [access token](https://developers.facebook.com/docs/facebook-login/access-tokens). If your app uses one of the Facebook SDKs, this token lasts for about 60 days. However, the SDKs automatically refresh the token whenever the person uses your app, so the tokens expire 60 days after last use. If your app doesn't use the Facebook SDKs, you need to include code that refreshes the User token manually. If the User access token expires, your app must put the user through the login flow again."

https://developers.facebook.com/docs/facebook-login/auth-vs-data

gustavlrsn (commented):
LGTM

Issue #640, opened 2022-03-25
[BUG] Minor Copy change: replace 'member' with 'participant' in message on round not participating in


eden-gt:
**Describe the bug**
The incorrect terminology is used in the info message displayed when a user who is not yet a participant in a round clicks on the 'participants' section of a round. 

**To Reproduce**

1. Go to a public round page (of a round you are not a participant in)
2. See info message of "You need to be a member of this round to see the member list."

**Expected behavior**
info message to read:
"You need to be a participant in this round to see the participants list."

**Screenshots**
<img width="1307" alt="Screenshot 2022-03-25 at 15 01 10" src="https://user-images.githubusercontent.com/94849002/160136505-2a8495ef-ee7e-414e-917f-765a19d858a1.png">

Powersource:
I don't think it is, since i think it was merged after staging was frozen

tomomiq:
Hmm okay, I'd like to avoid this confusion for the people responding to the #needs-testing issues - we are making an effort to stay on top of them :) 

What would be the best way to do that... a louder announcement about staging being frozen? Adding a step to update tags?

aerugo:
Hmm, I don’t understand though @Powersource - staging is not frozen now. You merged this to main yesterday - so it should be on staging?

Powersource:
Ah I thought we were aiming to do a prod deploy now on Thursday but we're not then?

aerugo:
No, because custom instances was not ready for testing yesterday it could not be tested properly. No release this week.

tomomiq:
Confirmed! 

![Uploading Screenshot 2022-04-09 at 16.00.37.png…]()

PR #644, opened 2022-03-27
fix: change member to participant

alinauroz:
for https://github.com/cobudget/cobudget/issues/640


Powersource (changes requested):
should be `participant in` not `of`

Powersource (approved):

Issue #648, opened 2022-03-29
[MINOR] Copy: replace "member" with "participant" in warning message when trying to delete round participant with balance >0 


eden-gt:
**Describe the idea**
As an admin, when trying to remove a participant who has a balance of more than 0 from a round, a warning message pops up when you hover over "delete". 

This currently reads: 
"You can only remove a round member with 0 balance" 

"member" should be replaced with "participant" - so the warning text should read:
"You can only remove a round participant with 0 balance" 

<img width="1313" alt="Screenshot 2022-03-29 at 16 16 21" src="https://user-images.githubusercontent.com/94849002/160633480-b4037bcb-be91-4676-952e-ac8d25f54551.png">

eden-gt:
fixed! thanks

PR #659, opened 2022-04-02
fix: change member to participant

alinauroz:
for https://github.com/cobudget/cobudget/issues/648


ui/server/graphql/schema/index.js
```diff
@@ -313,6 +313,8 @@ const schema = gql`
     groupMemberships: [GroupMember!]
     roundMemberships: [RoundMember!]
     avatar: String
+    facebookId: String
+    googleId: String
```
Powersource, line 317: these should not be added unless we plan on using it in the frontend

Issue #647, opened 2022-03-28
[BUG] Cannot add any bucket co-creators


tomomiq:
**Describe the bug**
Nothing happens when I type names to add a co-creator 

**To Reproduce**

1. Go to bucket page 
2. Click on the Edit button next to co-creator 
3. Start typing a name of someone that is part of this group 
4. Nothing happens 

<img width="535" alt="Screenshot 2022-03-28 at 16 01 22" src="https://user-images.githubusercontent.com/1287179/160415684-8a57b160-44f0-4254-a819-bb8024bd78ba.png">

**Expected behavior**

For an auto-complete to start. 

**Browser and device (please complete the following information):**

Checked on Chrome and Safari

tomomiq:
hope we can redesign this at some point! 
confirming it's working, though. 

<img width="1394" alt="Screenshot 2022-04-04 at 19 44 30" src="https://user-images.githubusercontent.com/1287179/161602044-1b4ec7a8-e5f7-43a6-aa46-2100ce4a6191.png">

PR #661, opened 2022-04-02
fix: make isApproved Boolean

alinauroz:
for https://github.com/cobudget/cobudget/issues/647


Powersource:
Wow graphql is sensitive sometimes. This seems to fix it, thanks

Issue #521, opened 2022-01-20
Use different slugs for the different tabs on the bucket page


Powersource:
Right now if I go to the comment tab and refresh the page I end up on the default Bucket tab again.

Thinking we should fix this by either appending something like `/comments` or `?tab=comments` to the url

francesca-sp:
Looks good!

PR #663, opened 2022-04-03
Bucket page slugs

alinauroz:
for https://github.com/cobudget/cobudget/issues/521

### DIFF
--- a/ui/pages/[group]/[round]/[bucket]/index.tsx
+++ b/ui/pages/[group]/[round]/[bucket]/index.tsx
@@ -116,7 +116,7 @@ const BucketIndex = ({ round, currentUser, currentGroup, router }) => {
 
   const tabsList = useMemo(() => ["bucket", "comments", "funders"], []);
   useEffect(() => {
-    const index = tabsList.findIndex(tab => tab === router.query.tab);
+    const index = tabsList.findIndex((tab) => tab === router.query.tab);
     setTab(index > -1 ? index : 0);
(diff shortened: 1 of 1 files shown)


ui/pages/[group]/[round]/[bucket]/index.tsx
```diff
@@ -107,13 +106,20 @@ const BucketIndex = ({ round, currentUser, currentGroup, router }) => {
...
     currentUser?.currentCollMember?.isApproved &&
     round.bucketReviewIsOpen &&
     round.guidelines.length > 0 &&
     bucket?.published;
 
+  const tabsList = useMemo(() => ["bucket", "comments", "funders"], []);
+  useEffect(() => {
+    const index = tabsList.findIndex(tab => tab === router.query.tab);
+    setTab(index > -1 ? index : 0);
+  }, [router.query.tab, tabsList]);
```
Comment (Powersource) [inline ui/pages/[group]/[round]/[bucket]/index.tsx:121]: When I've navigated through the tabs and then change the url by navigating through my history (`ctrl+left/right`) the ui doesn't update. I would think that having the `router.query.tab` here would be enough for that. I'll see if there's an easy fix.

Comment (Powersource) [reply ui/pages/[group]/[round]/[bucket]/index.tsx:121]: I went into a bit of a rabbit hole trying to improve this but the libraries we use were really not cooperating. Next.js isn't very well optimized for query param navigation I think, maybe having it as normal `/[param]` would work better, I think it did in Realities. But this is probably good enough for now so I'll merge.

Issue #513, opened 2022-01-17
[FEATURE] Invite links


gustavlrsn:
<img width="1116" alt="Screenshot 2022-04-07 at 11 51 36" src="https://user-images.githubusercontent.com/7785081/162176337-bdf60e39-00cb-4108-89ea-06dc406a24f6.png">

<img width="1093" alt="Screenshot 2022-04-07 at 12 04 04" src="https://user-images.githubusercontent.com/7785081/162176360-99ed1cf2-4812-4767-9538-be34266f4fe6.png">

- A new invite link is created with the "Create invite link" button. Before it is created, it does not exist.
- An invite link can be sent to existing cobudget users and to new users. If it is sent to new users, it first lets you register an account. See the regular invite user by email flow.
- If the invite link is deleted, it should stop working. If a new invite link is created, it has another url. This is to allow round owners to correct if an invite link has been sent to the wrong people by mistake.
- "Cancel" should be renamed to "Close" since it is otherwise confusing for a user who might think that their invite link is not saved if they "cancel" the prompt.

aerugo:
@alinauroz - testing this feature, and I get this error when trying to join by invite link on staging:
<img width="1380" alt="Screenshot 2022-04-28 at 12 16 19" src="https://user-images.githubusercontent.com/7785081/165732129-dfadcf1b-6539-4179-a10d-7667e2e868ab.png">

gustavlrsn:
Might be related to the type error I fixed that errored out the building of the project. Added `String()` around the inviteNonce because the sign function expected a `string`

gustavlrsn:
Updating the interface

PR #682, opened 2022-04-11
Feat/invite link

alinauroz:
for https://github.com/cobudget/cobudget/issues/513


alinauroz:
Only the admin should have permission to fetch, create and delete invitation links. 

The link was supposed to be saved in the database and it should not be visible to all users. Prisma does not allow to unselect a column by default. Instead, it requires to explicitly select or unselect column. The round table is read at multiple places in the code and at some place developer may forget to hide the invitation link.

To solve this problem, I did the following

`link = sign({roundId, nonce}, key)`

The nonce (time based) is generated when an admin requests to create new link. The nonce does not need to be hidden. I used jwt for signing. It has some more benefits. For example, the expiration time of an invitation link can be set without doing any change in the database.

ui/.env
```diff
@@ -1,3 +1,6 @@
 KEYCLOAK_CLIENT_ID="dreams"
 KEYCLOAK_REALM=plato
-DATABASE_URL='postgres://postgres:postgres@localhost:35432/cobudget'
\ No newline at end of file
+DATABASE_URL='postgres://postgres:postgres@localhost:35432/cobudget'
+
+#JWT_KEY
+JWT_KEY=&^13YG*&546TR*TYGE78*&@&5!)O@_I)12
```
Powersource, line 6: Shouldn't be here

ui/server/graphql/resolvers/index.ts
```diff
@@ -259,6 +261,30 @@ const resolvers = {
         return null;
       }
     },
+    roundInvitationLink: async (parent, { roundId }, { user }) => {
+      const isAdmin =
```
Powersource, line 265: Can we replace this with the normal isCollOrGroupAdmin?

ui/server/graphql/resolvers/index.ts
```diff
@@ -259,6 +261,30 @@ const resolvers = {
...
+        throw new Error("You need to be admin to fetch invitation link");
+      }
+
+      const round = await prisma.round.findFirst({
+        where: {
+          id: roundId,
+        },
+      });
+      return {
+        link:
```
Powersource, line 282: Why is this nested?

ui/server/prisma/schema.prisma
```diff
@@ -155,6 +155,8 @@ model Round {
   statusAccount   Account? @relation(fields: [statusAccountId], references: [id])
   statusAccountId String?  @unique
 
+  inviteNonce   Int?
```
Powersource, line 158: Is Nonce a good name here? My understanding is that you're only supposed to use a nonce ("number once") once and then throw it away. Maybe not that important of a difference though.

ui/pages/invite/[token].tsx
```diff
@@ -0,0 +1,80 @@
+import { useEffect } from "react";
+import { useRouter } from "next/router";
+import { useMutation, gql } from "urql";
+import HappySpinner from "components/HappySpinner";
+
+const JOIN_ROUND = gql`
+  mutation JoinRoundInvitationLink($token: String!) {
+    joinRoundInvitationLink(token: $token) {
+      id
+      round {
```
Comment (Powersource) [inline ui/pages/invite/[token].tsx:10]: Maybe should add the round id here, graphql tends to yell at me when I don't fetch that.

ui/server/graphql/resolvers/index.ts
```diff
@@ -710,6 +736,84 @@ const resolvers = {
...
+      const payload = verify(token);
+
+      if (!payload) {
+        throw new Error("Invalid invitation link");
+      }
+
+      const { roundId, nonce: inviteNonce } = payload;
+
+      const round = await prisma.round.findFirst({
+        where: { id: roundId, inviteNonce },
```
Powersource, line 794: Nice, this is pretty elegant

ui/components/InviteMembersModal.tsx
```diff
@@ -130,13 +178,60 @@ const InviteMembersModal = ({
...
-                Cancel
+                Close
+              </Button>
+              <Button
+                className="mr-2"
+                loading={createInviteLoading}
+                onClick={() => {
+                  createInviteLink({
+                    roundId,
+                  }).then(() => toast.success("Invite link created"));
```
Powersource, line 231: error handling

ui/components/InviteMembersModal.tsx
```diff
@@ -130,13 +178,60 @@ const InviteMembersModal = ({
...
+                      disabled: true,
+                      value: link,
+                    }}
+                  />
+                  <span className="mt-2 ml-2">
+                    <IconButton
+                      onClick={() => {
+                        deleteInviteLink({
+                          roundId,
+                        });
```
Powersource, line 208: error handling

Powersource (changes requested):
I read through the code now, looks pretty good but some minor changes needed.

Tomorrow I'll test the code as well.

@alinauroz do you want to fix the comments and the merge conflict? Otherwise I could also do that tomorrow probably.

ui/.env
alinauroz: should be in env.local, right?

Powersource: For the end-user yes (or configured in vercel in our case) but I would suggest just mentioning JWT_KEY in .env.local.default so that the end user knows that they need to set it. Also add instructions in either that file or in the readme telling the user how to generate it.

ui/server/prisma/schema.prisma
alinauroz: this number is only generated once. Once generated, this number can not be generated again for the same round (assuming max one link is generated in 1ms and clock not moved behind).

ui/.env
Powersource, line 6: Note to self to add this to vercel (different vars for prod and staging probably) before deploying

Powersource: We already have COOKIE_SECRET, we might be able to reuse that :thinking:

ui/server/prisma/schema.prisma
Powersource: I mean that to be called a cryptographic nonce https://en.wikipedia.org/wiki/Cryptographic_nonce I think we'd have to delete the nonce once we've sent it to a user or something, but here we might sent it to many many users. `inviteId` might make more sense in this case I think. No big deal but might make future developers less confused.

ui/server/graphql/resolvers/index.ts
alinauroz: A group admin who is not the member of a round in that group, should not be able to invite users.

ui/server/graphql/resolvers/index.ts
alinauroz: to match schema

ui/.env
alinauroz: yes, I think using COOKIE_SECRET is fine

ui/components/InviteMembersModal.tsx
```diff
@@ -228,7 +229,9 @@ const InviteMembersModal = ({
                 onClick={() => {
                   createInviteLink({
                     roundId,
-                  }).then(() => toast.success("Invite link created"));
+                  })
+                  .then(() => toast.success("Invite link created"))
+                  .catch(() => toast.error("Could not created invite link"));
```
Powersource, line 234: urql mutations don't reject https://formidable.com/open-source/urql/docs/basics/react-preact/#handling-mutation-errors

Powersource (changes requested):

package.json
```diff
@@ -29,5 +29,8 @@
     "npm-run-all": "^4.1.5",
     "prettier": "2.2.1",
     "typescript": "4.4.3"
+  },
+  "dependencies": {
+    "dotenv": "^16.0.0"
```
Powersource, line 34: uninstall

ui/server/graphql/resolvers/index.ts
Powersource: yeah i know i was wondering if we could simplify the schema. But no big deal, you can 'Resolve' this if you want.

ui/.env
Powersource: we're using `COOKIE_SECRET` instead

ui/server/graphql/resolvers/index.ts
alinauroz: I think it is ok to keep like this. If in future adding more field to this type is required, it would be easy to handle the changes.

Powersource:
UI looks good

ui/components/InviteMembersModal.tsx
Powersource: github just errors when i try to click resolve on this :upside_down_face:

alinauroz: On my account, github says it is resolved.

ui/server/prisma/schema.prisma
alinauroz: User is not supposed to work with the nonce. They will share the invite links. I intentionally named it nonce, so if this *number* generation logic is changed, developer knows that this number should not be repeated. I think it is ok to keep like this.

Issue #673, opened 2022-04-06
[FEATURE] Toggle direct funding for a round


aerugo:
![image](https://user-images.githubusercontent.com/7785081/162046446-40df16a5-8261-4162-aa0d-369f6d0e66aa.png)

- Add "Accept direct funding" to round funding settings page 
- If round does not belong to an org, show the button to set up with a stripe account
- Allow round admin to define terms. We might want to come up with a guide to instruct admins about what to consider.
- When direct funding has been set up and stripe has been connected, the direct funding options show up on all buckets in the round.

Powersource:
Hugi and I decided to start with Standard accounts (with Direct charges), with Express accounts as a potential future feature

Powersource:
The pr for this is merged now, but it's only in an integration branch so I'll keep it untagged as needs-testing for now

francesca-sp:
We tested this already last week, and it worked, so closing this. We added the extra issue about the feedback loops that is missing, here https://github.com/cobudget/cobudget/issues/672

PR #684, opened 2022-04-13
Toggle direct funding for a round

Powersource:
for https://github.com/cobudget/cobudget/issues/673

Adds an option in rounds to integrate with stripe

Note that this doesn't merge into main

* [x] split it into two modals instead (probably) one for integration, one for enabling and writing
* [x] make redirect back from stripe static but then redirect the user to where they were anyway
* [x] how exact does the list of allowed return urls in the stripe settings have to be? do we need to include the whole paths or is the domain enough?
* [x] fix slug navigation for settings so we can redirect back to the correct settings page
* [x] make the toggle and text forms also work
* [x] don't allow enabling if stripe isn't set up (in both api and ui)
* [x] update logic for detecting if the stripe integration is working (i.e. ping stripe)
* [x] make the forms look ok
* [x] add env vars to staging and prod
* [x] final test, go through the flow again, see if a half-filled-in account properly gets marked as not ready
* [x] fix crash on logged out

relevant documentation https://stripe.com/docs/connect/enable-payment-acceptance-guide


ui/server/graphql/resolvers/helpers.ts
```diff
@@ -324,3 +355,14 @@ export async function roundMemberBalance(member) {
...
   return totalAllocations - totalContributions;
 }
+
+/** only call this if you've verified the user is at least a round admin */
+export async function stripeIsConnected({ round }) {
+  if (!round.stripeAccountId) {
+    return false;
+  }
+
+  const account = await stripe.accounts.retrieve(round.stripeAccountId);
```
Powersource, line 365: Note that this will slightly slow down page loads for admins since this is called in `TOP_LEVEL_QUERY` and is an api call to stripe's servers. The slowdown might not be noticeable though. In the future if we want to speed it up we can move the parts of the graphql query that's relevant to the settings into the settings page.

Powersource: i ended up doing this anyway because of a permissions bug

Issue #683, opened 2022-04-11
[FEATURE] Toggle on and off admin specific notifications 


francesca-sp:
Add a section to the user email notification management section that shows admin specific notifications. 

This section only appears if you are an admin. For now, the first notification email toggle to add is: 
* A user requests to join a round

This turns on and off the emails that are sent when a user requests to join a group.

alinauroz:
Email settings are general. By general I mean they are not group or round specific. So, this settings will only appear if a user is an admin of at-least one round.

*What should happen in the following edge case?*
*A user is an admin of round and changes admin email settings. Later, he is no more an admin of any round and he wants to change/revert admin email settings.*

aerugo:
Good point @alinauroz. 

@francesca-sp - before we implement this we will need to look into round and group specific notifications.

francesca-sp:
Tested and is working as expected. At a later date we can do a bit of design work on the notifications page, but that seems out of scope for now.

PR #689, opened 2022-04-16
Feat/admin mail settings

alinauroz:
for https://github.com/cobudget/cobudget/issues/683

### DIFF
--- a/ui/server/services/EmailService/email.service.ts
+++ b/ui/server/services/EmailService/email.service.ts
@@ -55,7 +55,7 @@ export default {
   roundJoinRequest: async ({ round, roundMember }) => {
     const admins = await prisma.roundMember.findMany({
       where: { roundId: round.id, isAdmin: true },
-      include: { user: true },
+      include: { user: { include: { emailSettings: true } } },
     });
 
     const user = await prisma.user.findUnique({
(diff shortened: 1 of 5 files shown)


ui/server/services/EmailService/email.service.ts
```diff
@@ -68,7 +68,7 @@ export default {
 
     const roundLink = appLink(`/${group.slug}/${round.slug}`);
 
-    const adminEmails: SendEmailInput[] = admins.map((admin) => ({
+    const adminEmails: SendEmailInput[] = admins.map((admin) => admin.user?.emailSettings?.roundJoinRequest ? ({
```
Powersource, line 71: I have a memory that we should do it like `?? true` like in the other cases to catch the case where a user doesn't have `emailSettings` yet

Powersource (changes requested):
I'll fix this in a little bit

ui/server/services/EmailService/email.service.ts
alinauroz: I think user always have email settings. If a user didn't set email settings, postgres will return the default settings. I have tested a case, where user didnt have email settings. 

By default roundJoinRequest is true

Powersource: In the user model it's defined as `emailSettings    EmailSettings?`, so it shouldn't exist for a new user. When fetching it to render the settings page it's upserted so that might be why it worked for you, but I suspect it might crash for a user that's never visited the settings page.

Issue #213, opened 2021-04-20
[FEATURE] UI translation


skoteskote:
We need to implement localization of the frontend client. 
Because of overlapping communities, we would like to use the same approach as [OpenCollective](https://github.com/opencollective/opencollective-frontend) which uses [crowdin.com](https://crowdin.com/project/opencollective). 

Implementing localization will be done in steps:

- [x] Set up a project on Crowdin (https://crwd.in/cobudget)
- [x] Check with @znarf of OpenCollective if they are planning to stick with Crowdin. Done: They are happy with it and recommend it.
- [x] Research Crowdin, the OpenCollective implementation, and devise a plan for implementing Crowdin in Cobudget
- [x] Research what needs to be done to enable support for localization in RTL languages with Crowdin
- [x] Replace all hardcoded strings with localization library variables
- [x] Implement the first partial translation of a single page into Swedish
- [x] Test RTL support by implementing a partial Arabic translation of a single page

aerugo:
Closed in issue clean-up.

skoteskote:
I can produce the translated text files, just let me know how I should structure them.

skoteskote:
I would like to prioritize this issue so it is done by the summer, when I will start to reach out to orgs as a part of the Kulturbryggan project.

Powersource:
This is on staging now

francesca-sp:
Could you say more about what needs to be tested here from our side @Powersource ?

PR #698, opened 2022-04-25
Feat/translation

alinauroz:
for https://github.com/cobudget/cobudget/issues/213


ui/components/Layout.js
```diff
@@ -37,17 +42,36 @@ const Layout = ({
...
           {process.env.TERMS_URL && (
             <LinkOut href="/terms-and-conditions">Terms and Conditions</LinkOut>
           )}
+          <select onChange={(e) => changeLocale(e.target.value)}>
+            {
+              supportedLangs.map((option) => (
+                <option 
+                  key={option.value} 
+                  value={option.value}
+                  selected={locale === option.value}
```
Powersource, line 68: Getting this warning in the console

```
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
```

ui/components/Layout.js
```diff
@@ -37,17 +42,36 @@ const Layout = ({
...
           {process.env.PRIVACY_POLICY_URL && (
-            <LinkOut href="/privacy-policy">Privacy Policy</LinkOut>
+            <LinkOut href="/privacy-policy">
+              Privacy Policy
+            </LinkOut>
           )}
           {process.env.TERMS_URL && (
             <LinkOut href="/terms-and-conditions">Terms and Conditions</LinkOut>
           )}
+          <select onChange={(e) => changeLocale(e.target.value)}>
```
Powersource, line 62: I'm unable to click the dropdown. I think my resolution/zoom makes the page too small

image

Powersource: This probably only applies to the bucket page

ui/lang/sv-SE.json
```diff
@@ -0,0 +1,44 @@
...
+  "HSBYFr": "allowed.",
+  "J2D0+i": "Mark as completed",
+  "Jltspt": "+ Bilder",
+  "K3r6DQ": "Delete",
+  "Lvy00l": "Existing funds",
+  "MJ2jZQ": "Total",
+  "NYQP6J": "stretch goal:",
+  "NaGSue": "Max.",
+  "OXKkW2": "Yes, cancel it",
+  "Q6iAKB": "Are you sure you want to cancel this {bucketName}?",
```
Powersource, line 22: When we translate this and similar cases, I wonder how we should handle the names we get from env vars. One env var per word and language can't work. Maybe the most reasonable solution is to only allow custom nouns when the user is on english? @aerugo

Powersource (changes requested):
A few cases don't have the translation component yet, at least the "New round" and "New bucket" buttons

Powersource:
This doesn't include translation of emails, which I think is fine for now but we shouldn't close the issue before that's also done in another pr

aerugo:
Since we don't save language preferences for users (it's just stored in a cookie) we would need to reconsider that if we were to also translate emails.

I suggest we limit this issue to the browser client UI and open a separate issue for email translation.

ui/.babelrc
```diff
@@ -0,0 +1,12 @@
+{
```
Powersource, line 1: I got this when compiling

```
info  - Disabled SWC as replacement for Babel because of custom Babel configuration ".babelrc" https://nextjs.org/docs/messages/swc-disabled
info  - `compiler` options in `next.config.js` will be ignored while using Babel https://next.js.org/docs/messages/ignored-compiler-options
info  - Using external babel configuration from /home/me/prj/plato/cobudget/ui/.babelrc
```

which makes us lose the styled components optimization, and I think eventually this will be deprecated. Can we avoid adding this file?

ui/lang/en.json
```diff
@@ -0,0 +1,44 @@
+{
```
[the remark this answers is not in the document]
alinauroz: Yes, this file should be there. Translations are done on basis of this file.

We are using crowdin + react-intl for translations. It works like this

1. Text is rendered using FormatMessage component which is imported from react-intl. The default message is the message which appears if a translation is missing.
2. The id of the text is generated using hash function.
3. `yarn extract` command extracts the default messages from the code and assign them an id. These are stored in en.json file.
4. The en.json file is uploaded to crowdin using `yarn crowdin:upload` command.
5. After translating text, translations can be downloaded using the command `yarn crowdin:download`

ui/lang/sv-SE.json
```diff
@@ -0,0 +1,44 @@
+{
+  "/AmWqb": "+ Omslagsbild",
+  "0KKXrH": "Budget",
+  "26gDq9": "Finansieringsmål:",
+  "3g96gM": "Inga bidrag än",
+  "47FYwb": "Cancel",
+  "4U2uvI": "Completed on {day}",
```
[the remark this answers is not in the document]
alinauroz: The translation files are not supposed to be edited manually. Crowdin handles the translations.

ui/package.json
```diff
@@ -13,13 +13,18 @@
     "generate": "yarn run generate:prisma",
     "generate:prisma": "prisma generate",
     "migrate": "npx prisma migrate dev",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "tsc --noEmit",
+    "extract": "formatjs extract 'components/**/*.*s*' --ignore='**/*.d.ts' --out-file lang/en.json --id-interpolation-pattern '[sha512:contenthash:base64:6]' --format ../scripts/formatter.js",
+    "crowdin": "crowdin",
+    "crowdin:download": "../scripts/crowdin.sh download",
```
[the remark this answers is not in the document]
alinauroz: 2 env variables are required. Samples are there in .env.local.default file.

ui/lang/sv-SE.json
alinauroz: fixed

ui/components/Layout.js
alinauroz: fixed

ui/.babelrc
Powersource: In the discussions in the next.js repo there seem to be a bunch of users requesting formatjs support, so that will probably get in eventually. We can keep this file for now, but we should re-add the styled-components config that we had before into it so we don't miss that optimization https://github.com/cobudget/cobudget/commit/7e66cd7f50595250746ae7bea05189e0928c9bae

alinauroz: Alright, Ill keep checking

Powersource:
A dependency of a dependency started using react types from react 18 so had to pin it using `resolutions` https://stackoverflow.com/questions/71791347/npm-package-cannot-be-used-as-a-jsx-component-type-errors/71828113#71828113

PR #699, opened 2022-04-29
Direct funding of buckets

Powersource:
Integration branch for this milestone https://github.com/cobudget/cobudget/milestone/15

* https://github.com/cobudget/cobudget/pull/684
* https://github.com/cobudget/cobudget/pull/696


Powersource:
We're currently connecting it all to the greaterthan stripe. I suspect that once the cobudget association has a stripe and we want to start using that one, businesses that are already connected will have to reconnect with a technically new connected stripe account. And we'll then have to clear old account ids from the db, since we won't be their platform anymore. This would probably also break refunds through the app, but maybe the few cases where that would happen could be manually handled by the businesses.

aerugo:
Good point @Powersource, we need to keep an eye on that. We should have the new account sometime in the fall.

PR #700, opened 2022-05-02
Fix/loading issues

alinauroz:
1. Fixes the bug where Login and Signup appeared in the top bar even if the user was logged in.
2. Fixes the bug on bucket page. when the page is loaded, an error appears for short time saying the bucket does not exist or you do not have access to it. Check bug in the screenshot.

Screen Shot 2022-05-02 at 5 35 30 AM

### DIFF
--- a/ui/components/Header/index.tsx
+++ b/ui/components/Header/index.tsx
@@ -9,6 +9,7 @@ import toast from "react-hot-toast";
 import Link from "next/link";
 import { useRouter } from "next/router";
 import Head from "next/head";
+import { LoaderIcon } from "components/Icons"
 
 const css = {
   mobileProfileItem:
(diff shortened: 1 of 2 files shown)


ui/components/Header/index.tsx
```diff
@@ -102,13 +104,31 @@ const Header = ({
...
+    const user = get("user");
+    if (user && fetchingUser) {
+      setUserFromStorage(user);
+    }
+    else if (!fetchingUser && currentUser === null) {
+      remove("user");
+      setUserFromStorage(null);
+    }
+  }, [fetchingUser, currentUser]);
+
```
gustavlrsn, line 131: Instead of storing the user in local storage, would it be possible to decode the cookie already saved instead?

gustavlrsn (commented):
The issue on loading the bucket pages is now fixed by not doing static generation, so I think we might want to skip that fix.

Regarding storing the user in local storage, I think it should be possible to use the cookie directly instead, but I might be overlooking something

Issue #702, opened 2022-05-02
[FEATURE] "A new bucket is published" default to off in email notifications


skoteskote:
**Describe the feature**
The "A new bucket is published" toggle should be off by default in email notifications. The email notifications becomes very spammy when it is on and the platform is used for a workshop. 

Context: MT recently did a workshop where 30 ideas were added during the course of 30 minutes. Each user subsequently got 30 emails informing them of this, leading to a bad first experience with the otherwise very useful email notifications feature.

Powersource:
related https://github.com/cobudget/cobudget/issues/622

Powersource:
This is on staging now

Powersource:
It will be off for new users, including users who haven't visited their email settings page

francesca-sp:
tested, and works correctly!

PR #711, opened 2022-05-08
Bucket publish mail settings

alinauroz:
for https://github.com/cobudget/cobudget/issues/702 and a typo fix. Bucket should be singular.

Screen Shot 2022-05-09 at 2 34 40 AM


Powersource:
We also need to change the default in `email.service.ts`, I'll do that tomorrow

alinauroz:
> We also need to change the default in `email.service.ts`, I'll do that tomorrow

Check, I have done. But I don't think it was required because email settings are always there for a user.

Powersource:
Last time we had a PR related to this I tested it, and users don't have the settings until they've visited the email settings page (which is probably most users).

Issue #678, opened 2022-04-06
[FEATURE] Direct fund a bucket (logged in user)


aerugo:
![image](https://user-images.githubusercontent.com/7785081/162051251-0757b0c2-4d79-4bf2-9da9-5b7ca3e4d550.png)

- Fund pane should show two options - "From my balance" and "Direct with card"
- If user balance is more than 0, preselect "From my balance" otherwise preselect "Direct with card"
- If funding is a transaction (see #674) show terms and VAT
- If bucket owner has set a minimum amount, restrict input of lower amounts
- Show a contribution summary before the payment button

We use the Stripe Connect flow for these payments, see Stripe documentation.

gustavlrsn:
What we discussed earlier I think was that to allow a non-member to fund a bucket, we want to have the Contribution, Allocation and Transaction schemas rely on `userId`'s instead of `roundMemberId`'s, (instead of creating dummy roundMembers). 

This would also require some changes in adding an `incomingAccount` to the `User`, to have a `fromAccount` to specify in the Transaction. I think it could make sense to have all incoming transactions tied to one account on the user instead of on multiple accounts on the round members? The incoming account becomes a tally of everything I as a person/user has put into the platform, it seems to make more sense and be simpler to have this tied to my user account.. I think..

Powersource:
@gustavlrsn Do you have recommendations on how to do those migrations? Something similar to https://github.com/cobudget/cobudget/issues/578 ?

gustavlrsn:
@Powersource since we now have multiple instances to keep track on, and there could potentially be other instances out there where the migration should be automatic, ideally the complete migration should to be part of the sql migration script generated by prisma (modified by us to do the data migrations).

The other approach with running javascript files with prisma code is hard to do automatically in unison with the other migrations that are applied, I think. Even if we found a way to run them in unison, the prisma client gets "outdated" with old types as the schema changes, and making it tricky to build and run the prisma client in a previous step.

**_But_**, since I guess we can be pretty sure no one else is using a custom instance yet, we could also write javascript migrations with prisma commands that we apply manually to the different dbs, running them on your local machine and just swapping out the database url, if it seems too cumbersome to figure out the correct SQL operations needed. Working with a typed prisma client is pretty convenient compared to writing raw SQL after all..

Powersource:
This pattern is interesting https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern

[...]

PR #720, opened 2022-05-11
Allow logged in users to directly fund buckets

Powersource:
For https://github.com/cobudget/cobudget/issues/678

Merging into this integration branch https://github.com/cobudget/cobudget/pull/699


ui/pages/api/stripe/webhook.ts
```diff
@@ -0,0 +1,69 @@
...
+    const bucket = await prisma.bucket.findUnique({
+      where: { id: metadata.bucketId },
+    });
+
+    const missing = () => {
+      throw new Error(`Missing metadata in session ${session.id}`);
+    };
+
+    await prisma.$transaction([
+      prisma.transaction.create({
```
Powersource, line 51: In `ui/server/controller/index.ts` where we `Allocate` to the user normally using the `allocateToMember` function, we also have another `Transaction`. Maybe we could make use of that function instead of making new allocation code here.

ui/pages/api/stripe/create-checkout-session.ts
```diff
@@ -0,0 +1,168 @@
...
+  bucket: Bucket;
+  roundMember: RoundMember & { user: User };
+}): Promise<string[]> {
+  if (bucket.directFundingType === "DONATION") return [];
+
+  const taxRates = await stripe.taxRates.list({ limit: 100 });
+
+  //TODO
+  if (taxRates.has_more)
+    throw new Error("We should implement pagination for taxes");
```
Powersource, line 23: TODO pagination for tax rates

gustavlrsn (approved):
Skimmed it and it looks good to me!

Issue #666, opened 2022-04-04
[FEATURE] re-direct to the last visited Org page after logging in 


tomomiq:
**Describe the feature**

Currently, if you are part of multiple organizations or SoloRounds, you are taken to this page after logging in - which is confusing.   
How about if you get re-directed to the Org Top (or SoloRound page) of whatever page you were last on? And if you just created your account, re-direct to /new-round. 

<img width="1365" alt="Screenshot 2022-04-04 at 22 24 33" src="https://user-images.githubusercontent.com/1287179/161626147-be4561b4-1a9b-4525-bebc-5a9a7bae568a.png">

alinauroz:
Should we save users last active page and update it on every page load?

aerugo:
Yes, I guess we can do that. In a cookie, I guess? Not need to save it in the database, I think.

skoteskote:
If we do it in a cookie and we want to save it after the user session we need cookie approval banner and policy.

gustavlrsn:
I think a common approach is to save the redirect path in a query parameter in the URL

gustavlrsn:
Basically take the url you were on before clicking Login

alinauroz:
> I think a common approach is to save the redirect path in a query parameter in the redirect URI

I add a `r` param in url to redirect. 

But for example when a user has closed browser and he opens the website again (cobudget.com), we do not know the where he was last time.

aerugo:
@alinauroz - can you describe a case where we need to know where a user was before they closed a browser window?

alinauroz:
> @alinauroz - can you describe a case where we need to know where a user was before they closed a browser window?

Im saving the latest page's pathname in localStorage every-time a user visits any page on cobudget and if user is logged in.

aerugo:
I understand. But this is my question:
When do we need that? What is the user story when we need to remember where a user was if they close their browser, instead of just storing the redirect path in a url parameter like Gustav suggests?

I am asking these clarifying questions to understand if you might have thought of some user story where this is needed, that I might not be aware of.

alinauroz:
> I understand. But this is my question: When do we need that? What is the user story when we need to remember where a user was if they close their browser, instead of just storing the redirect path in a url parameter like Gustav suggests?
> 
> I am asking these clarifying questions to understand if you might have thought of some user story where this is needed, that I might not be aware of.

[...]

PR #721, opened 2022-05-11
redirect to last visited page after login

alinauroz:


alinauroz:
alternative fix for https://github.com/cobudget/cobudget/issues/666

closed PR# https://github.com/cobudget/cobudget/pull/709

Issue #738, opened 2022-05-30
[BUG] All kinds of characters in usernames


gustavlrsn:
**Describe the bug**
You can basically set your username to anything you want, including spaces, funky characters and what not.

The username should ideally be something simple to type (for the @at functionality etc) and be URL-friendly.

**Suggestion of fix**
- [x] Allow `A-z` `0-9` `-`, (maybe `.`?) and no special characters. Make sure we have case insensitive uniqueness.
- [x] When we check if a user has a username (and open the "complete signup" modal), we can also check if they have a bad username and ask them to fix it, showing a different modal or at least a different wording.

This will also fix the issue of emails as usernames form Keycloak.

Powersource:
This is a good fix. In the future would be nice to allow non-latin names as well.

tomomiq:
LGTM!

PR #749, opened 2022-06-02
Make users accept updated terms and keep track of version

### LINKED ISSUE #500: [FEATURE] Keep track of which version of the ToS a user has accepted
To update terms of service we should keep track of which version of the ToS a user has accepted.

gustavlrsn:
Fixes #500.

Adding a `RequiredActions` component which can handle all required actions a user might have to do when logging in:
- finishing signup (setting name, username, accepting terms)
- accepting updated terms
- and things like fixing faulty username #738


gustavlrsn:
This is ready for review. I also added the fix username prompt to the Required actions component, and updated the legacy edit profile modal to be of new style and checking case insensitive uniqueness of usernames

Issue #756, opened 2022-06-06
[BUG] Bulk allocation times out 


gustavlrsn:
**Describe the bug**
When bulk allocation to thousands of members, we are using `allocateToMember` for each member, either in a loop or in parallell. The serverless function can either time out (if we are allocating in a loop), or it can run out of database connections (if we are parallelizing the requests).

**Proposal**

What would be better than to reuse is if we can do bulk allocation through its own function, where we only do a single `findMany` for all members and their transaction data, do the calculations for all before we do a single transaction with `createMany` requests to create all the things, as well as a single request to Postmark with a batch email sending request.

PR #759, opened 2022-06-08
Bulk allocate optimization

gustavlrsn:
Fix for #756.

Instead of doing `n * k` db queries when bulk allocating we are simply doing `k`


Powersource:
Linting!

gustavlrsn:
@Powersource added pre batching before using the postmark batch api call

Issue #371, opened 2021-11-05
[FEATURE] Show bucket status clearly on bucket page


aerugo:
As a user, I want to clearly see which part of the "life-cycle" a bucket is in when I am viewing the bucket.

francesca-sp:
There is feedback in #385 about the design and copy of this, which applies the same for this issue. This needs to be incorporated before it can be pushed live.

tomomiq:
Still waiting for labels to have the right colors :)

aerugo:
Closing this one and moving colors to new issue https://github.com/cobudget/cobudget/issues/814

PR #766, opened 2022-06-12
show bucket status on bucket page

alinauroz:
for https://github.com/cobudget/cobudget/issues/371


gustavlrsn (approved):
LGTM. We might want to update the styling of this but let's do that in a separate issue.

Issue #744, opened 2022-06-01
[FEATURE] Add progress bar to languages


aerugo:
We should add translation completeness progress bars to the languages, like on OpenCollective. Right now it feels buggy as  choosing a translation with only a few strings translated usually has no visible effect.

DavidsBees:
I cant find that progress bar - I see the translations here so far: 
![image](https://user-images.githubusercontent.com/80971095/177329939-f49851d2-a3d3-4fda-8eec-79313640af01.png)

Powersource:
Yeah this isn't merged to far so it's just been tagged wrong

Powersource:
If the relevant env vars are missing (they were on staging and prod, i've added them now. they're still missing on the custom domain deployments) then the progress shows as 0%. In that case it's better to hide the percentage.

PR #775, opened 2022-06-19
language progres

alinauroz:
for https://github.com/cobudget/cobudget/issues/744


package.json
```diff
@@ -30,5 +30,7 @@
     "prettier": "2.2.1",
     "typescript": "4.4.3"
   },
-  "dependencies": {}
+  "dependencies": {
+    "node-fetch": "^3.2.6"
```
Powersource, line 34: I think this should be in the other package.json?

PR #780, opened 2022-06-27
Add paid plans

### LINKED ISSUE #395: [FEATURE] Org can subscribe to CoBudget paid plan
As Cobudget, we want certain features to be limited to paid plans, and we want those to be enabled automatically when a user signs up for a paid plan with their organization. 

* [ ] Maybe add a tax_rate when checking out?

gustavlrsn:
Fixes #395


gustavlrsn:
Enables a signup flow for groups to that includes setting up monthly or yearly billing

ui/pages/api/stripe/create-checkout-session.ts
```diff
@@ -66,118 +69,165 @@ async function getTaxRates({
...
+        metadata: {
+          userId: req.user.id,
+          groupSlug: req.query.groupSlug,
+          groupName: req.query.groupName,
+        },
+        allow_promotion_codes: true,
+        ...customerMetadata,
+        billing_address_collection: "auto",
+        success_url: `${origin}/new-group/?upgraded=true&group=${slugify(req.query.groupSlug)}`,
+        cancel_url: `${origin}/new-group/?upgraded=false`,
```
Powersource, line 109: I don't think we can redirect like this, it has to be a static url. See the example lower down in this file on how to do it instead

ui/pages/api/stripe/create-stripe-billing-portal-url.ts
```diff
@@ -0,0 +1,34 @@
...
+        where: {
+            groupId,
+            userId: req.user.id
+        }
+    });
+    if (!groupMember.isAdmin) throw new Error("You need to be admin to access billing portal");
+
+    const { url } = await stripe.billingPortal.sessions.create({
+        customer: group.stripeCustomerId,
+        return_url: `${origin}/${group.slug}/settings/billing`,
```
Powersource, line 28: Can't redirect like this

ui/pages/api/stripe/create-checkout-session.ts
```diff
@@ -66,118 +69,165 @@ async function getTaxRates({
...
+      customer_email: req.user.email,
+    };
+    const origin = getRequestOrigin(req);
+    try {
+      const session = await stripe.checkout.sessions.create({
+        mode: "subscription",
+        payment_method_types: ["card"],
+        tax_id_collection: {
+          enabled: true,
+        },
```
Powersource, line 93: Is this needed?

ui/pages/api/stripe/create-checkout-session.ts
gustavlrsn: Hm, it is working in development. Any reasons why it should not work?

ui/pages/api/stripe/create-checkout-session.ts
gustavlrsn: It is to collect VAT ID, I guess many orgs using it will be companies/associations rather than consumers.

ui/pages/api/stripe/create-stripe-billing-portal-url.ts
gustavlrsn: This also works in development, and what would be another strategy?

ui/pages/api/stripe/create-checkout-session.ts
Powersource: I think in prod they might be stricter. I'll try to find the docs for it

ui/pages/api/stripe/create-checkout-session.ts
Powersource: Ah wait I was thinking this was a donation for some reason. But since they're paying us, shouldn't we add a tax rate like https://github.com/cobudget/cobudget/blob/91ca4288d97de0f333d4853651cc7e843c3727ca/ui/pages/api/stripe/create-checkout-session.ts#L133

ui/pages/api/stripe/create-stripe-billing-portal-url.ts
Powersource: The alternative would be like this https://github.com/cobudget/cobudget/blob/91ca4288d97de0f333d4853651cc7e843c3727ca/ui/pages/api/stripe/create-checkout-session.ts#L112-L116

Powersource: That saves it in a cookie that then gets used once you're back

ui/pages/api/stripe/create-checkout-session.ts
Powersource: I'll merge this anyway but add a note in the issue that we might wanna add this before closing

Issue #747, opened 2022-06-01
[FEATURE] When going "back" to bucket list from bucket, got to same page 


aerugo:
As people explore different buckets to fund, a likely behaviour will be to open buckets and then press "back" to find other buckets. Currently, this means that you do not end up on the same page as you were, and you need to scroll through again. This is especially annoying in large rounds, like the Borderland. 

We should save the page position for the user so they can navigate back to where they were in the bucket list.

gustavlrsn:
Maybe a query variable `pages` that keeps track of the number of pages you have loaded should be enough, I think Next.js will keep track of scroll position beyond that. So that from the second page you would have the url look like: `https://cobudget.com/c/your-round?pages=2`

Powersource:
Shouldn't urql automatically grab the data from the cache? It used to I think. Feels like we accidentally clear it somehow.

gustavlrsn:
It does but the number of pages loaded is a local state on the bucket page list, which is reset when the page is unmounted, so it should just be a matter of preserving that and then the urql cache will do the rest.

Powersource:
Hmm ok so I guess this is one of the drawbacks of this pagination strategy

DavidsBees:
It is not working reliably - i get randomly put back to the same location on the page OR at the top of the page. This seems to be independent from the bucket or its location on the page

aerugo:
@alinauroz - see comment from @DavidsBees

PR #781, opened 2022-06-27
Fix/same page use cache

alinauroz:
for https://github.com/cobudget/cobudget/issues/747


gustavlrsn:
Great that it works! One request that I h but could you change the offset hash to a query variable called `pages` where we keep track of the number of pages instead of the total offset, as [per this comment](https://github.com/cobudget/cobudget/issues/747#issuecomment-1144757704). I think that would be a cleaner solution, easier to use in the code and easier to understand as a user.

alinauroz:
@gustavlrsn Should we we save page number in url? If a user is on 20th page and share that link of the webpage with a group of people, everybody who opens this link will load 20 pages at once.

I have added a solution(https://github.com/cobudget/cobudget/pull/781/commits/2883c560dd40f0d7da7e8d8f1178b55699043749) which hides no of items loaded in url.

PR #812, opened 2022-08-23
always show all statuses

alinauroz:
I haven't linted the code. I think it may result in conflicts.


Powersource:
@alinauroz which issue is this for?

Powersource:
from discord

> nauroz: When there are no buckets in a round, there are no options in filter status. Same implementation is on cobudget.com.
> 
> I have modified to show status options even if they are 0. Is it ok? 
> 
> image
>
> 
> hugi: Yes
> hugi: That’s good.

PR #816, opened 2022-08-30
feat: error reporting

alinauroz:


ui/utils/reportError.ts
```diff
@@ -0,0 +1,29 @@
...
+    const fd = new FormData();
+    fd.append("content", "APP CRASH NOTIFICATION");
+    fd.append(
+      "file",
+      new Blob([content], { type: "text/plain" }),
+      "log-" + Date.now() + ".json"
+    );
+    fd.append(
+      "avatar_url",
+      "https://media3.giphy.com/media/cLqxgg4nke0iu8UpzD/giphy.gif"
```
Powersource, line 21: if we're allowed to use this image then download it and use directly instead

ui/.env.local.default
```diff
@@ -24,3 +24,4 @@ FROM_EMAIL="fake@example.com"
 #STRIPE_YEARLY_PLAN_PRICE_ID
 
 #TERMS_UPDATED_AT="2022-05-30T20:18:35.225Z"
+#ERROR_REPORTING_WEBHOOK=
```
Powersource, line 27: what should we set this to in prod?

Powersource:
@alinauroz which issue is this for?

aerugo:
> @alinauroz which issue is this for?

This has not been tracked in a corresponding issue. @alinauroz suggested working on this in a weekly tech meeting and we agreed.

ui/.env.local.default
alinauroz: A new webhook should be created on discord. In case of no webhook, there will be no error message in discord

aerugo:
@alinauroz conflicts here too

ui/.env.local.default
Powersource: Why on discord? Wouldn't that very quickly become unmanageable? But I guess we could maybe try it as a first iteration

Powersource: added now

Powersource:
Added localizations for the new strings but ran into this https://github.com/cobudget/cobudget/issues/851

ui/utils/reportError.ts
```diff
@@ -0,0 +1,26 @@
...
+    const fd = new FormData();
+    fd.append("content", "APP CRASH NOTIFICATION");
+    fd.append(
+      "file",
+      new Blob([content], { type: "text/plain" }),
+      "log-" + Date.now() + ".json"
+    );
+    fd.append("avatar_url", "https://cobudget.com/favicon.ico");
+
+    fetch(hookUrl, {
```
Powersource, line 21: @alinauroz I get these errors in my browser console when I cause this to run

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://discord.com/api/webhooks/1025000522022920192/1xjiaJ86fl2EzhJoUOxNwOdW1kIeRM2Krgql1fKlmUH9WJ62zAt11QN4xdEoU93N_jzW. (Reason: CORS request did not succeed). Status code: (null).

Uncaught (in promise) TypeError: NetworkError when attempting to fetch resource.
```

ui/utils/reportError.ts
```diff
@@ -0,0 +1,26 @@
+export default function reportError(error: Error) {
+  const { message, stack } = error;
+  const hookUrl = process.env.ERROR_REPORTING_WEBHOOK;
+  if (hookUrl) {
+    const content = JSON.stringify({
+      error,
+      message,
+      stack,
+      url: window?.location,
+    });
```
Powersource, line 10: If we do `JSON.stringify(..., nulll, 2)` then it'll be easier to read on discord

ui/utils/reportError.ts
Powersource: simply removed it

Issue #708, opened 2022-05-06
[FEATURE] Create new role: Super-User


DavidsBees:
**What?** 
A new role / access feature to be able to see all rounds / groups and its contents

**Why?**
By having support team members to be able to actually experience the environment of where the request is coming from, it is easier to solve issues immediately.
This is likely to decrease next level service requests of dev team. 

**Scope**
Seeing & acting in a group / round from an admin perspective
Seeing & acting in a group / round from a user perspective (could be deprioritized for now)

**Specifications:**
- In order to reduce risk, the super user should be be a separate account to login to OR anyone with super admin status has to explicitly start an super admin session when they want to do super user stuff
- Ideally there is a visual signal that makes it obvious to the person using that account that the super user is currently activated
- Super user changes are logged in a special log

PR #831, opened 2022-09-13
Feat/superadmin

alinauroz:


Powersource:
@alinauroz do we have an issue for this?

Powersource:
i know we want to add this but i want to know roughly what we're aiming for to make it easier for me to merge

aerugo:
> i know we want to add this but i want to know roughly what we're aiming for to make it easier for me to merge

It is described in #708

alinauroz:
@Powersource  the flow for superadmin is:

1. A new isSuperAdmin field is added to user which is false by default.
2. To make someone a superadmin, this field should be set to true by updating db
3. A superadmin needs to start a superadmin session to work as a superadmin
4. A superadmin can request 15, 30, 45 and 60 minutes superadmin session. A jwt is created with requested time and this is stored in HTTP only cookie.
5. A superadmin can end superadmin session before automatic session expiration
6. Superadmin sessions are also stored in db which tells when and who started a superadmin session for how much duration. The sessions can be seen on <domain>/ss

Powersource:
@aerugo @DavidsBees the spec says "Super user changes are logged in a special log". This PR logs when superadmin sessions are started but not what the admin does. I suspect that logging that would take a lot more code though. I'm leaning towards merging this PR being safe enough to try, you agree?

ui/components/Header/index.tsx
```diff
@@ -94,17 +96,72 @@ const JOIN_ROUND_MUTATION = gql`
...
   const [, joinRound] = useMutation(JOIN_ROUND_MUTATION);
   const color = round?.color ?? "anthracit";
+  const [superAdminTime, setSuperAdminTime] = useState<HTMLElement>();
+  const [inSession, setInSession] = useState(false);
+
+  useEffect(() => {
+    if (ss?.start + ss?.duration * 60000 - Date.now() > 0) {
+      setInSession(true);
+    }
+    if (superAdminTime && superAdminTime) {
```
Powersource, line 149: ?

ui/components/Header/index.tsx
```diff
@@ -94,17 +96,72 @@ const JOIN_ROUND_MUTATION = gql`
...
+      const interval = setInterval(() => {
+        const diff = ss?.start + ss?.duration * 60000 - Date.now();
+        if (diff < 0 || !ss) {
+          clearInterval(interval);
+          window.alert("Session Expired");
+          setInSession(false);
+          return;
+        }
+        if (superAdminTime && superAdminTime?.innerHTML) {
+          superAdminTime.innerHTML = toMS(diff) + "";
```
Powersource, line 159: Isn't it better to update a variable that react then renders?

ui/pages/_app.tsx
```diff
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react";
```
Powersource, line 1: A superadmin doesn't seem to be able to create a new bucket

ui/pages/ss.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <div className="page">
+      <div className="bg-white rounded-lg shadow overflow-hidden">
+        <TableContainer>
+          <Table aria-label="simple table">
+            <TableHead>
+              <TableRow>
+                <TableCell>Admin</TableCell>
+                <TableCell>Start</TableCell>
+                <TableCell>End</TableCell>
+                <TableCell>Duration</TableCell>
```
Powersource, line 68: The duration for the current session is a bit buggy, for a 15 min session it's listed as "15:00 / 15"

ui/pages/_app.tsx
aerugo: This is intentional, and the solution is to first invite yourself to the round. That way it will be less confusing for the group admins and participants when a new bucket appears.

ui/components/Header/index.tsx
aerugo: @alinauroz - this is the timeout, right? Please try to make the variables a little more legible.

alinauroz: Not timeout. superAdminTime is the reference of the element which shows superAdmin's remaining time on frontend. But it is only enough to it once.

ui/components/Header/index.tsx
alinauroz: Im updating after every second. I was concerned that there would be too much re-renders.

Powersource:
@alinauroz 

> I think we can also log admin activity easily. We can do something like this:

Yeah that's smart, I think we should do that

Powersource:
Made an issue for that https://github.com/cobudget/cobudget/issues/850

Issue #872, opened 2022-10-13
[BUG] Round funding information disappears when a deadline for submitting buckets is set


francesca-sp:
**Describe the bug**
When activating the setting of buckets having a deadline after which they can no longer be created, the whole info box about funds in a round disappears as well. This is not supposed to be the case, because only the "create bucket" button should disappear. 

**To Reproduce**

1. Go to any Cobudget round and activate in "Settings" > Funding > bucket creation closes > add a date there

<img width="919" alt="image" src="https://user-images.githubusercontent.com/3641828/195548955-83a59964-0ea9-4e75-aee9-10315f8078a1.png">

3. Go back to the main round overview page. 
4. You will see that both the create bucket and the overview of funds in the round and personal funds have disappeared. 

![image](https://user-images.githubusercontent.com/3641828/195549361-3dd50603-9e22-4371-809d-01268a012a2f.png)

**Expected behavior**
The expected behavior is that when a bucket creation deadline is activated, only the create bucket button disappears, and nothing else.

alinauroz:
@francesca-sp should this box be hidden when user is not the member of the round?

aerugo:
@alinauroz - the box should be publicly visible.

aerugo:
But if the user is not a member or if there is no user logged in, hide “ In your account” row.

PR #873, opened 2022-10-13
Fix/always show funding status

alinauroz:
for #872


alinauroz:
@aerugo this is ready

PR #896, opened 2022-11-20
E2e tests

alinauroz:
_Few tests' responses are stochastic. This needs to be fixed before merging this code to main._


alinauroz:
This MR is ready

cypress/e2e/bucket/filter-bucket.cy.js
```diff
@@ -0,0 +1,104 @@
...
+    createRound(roundSlug);
+  });
+
+  it("filters buckets on based on statuses", () => {
+    // Only 1 cancelled bucket is required
+    let cancelled = false;
+    for (let i = 0; i <= 4; i++) {
+      createBucket(roundSlug, `Bucket ${now}`);
+
+      cy.wait(10000);
```
Powersource, line 23: we should avoid stuff like this. https://filiphric.com/waiting-in-cypress-and-how-to-avoid-it

but fine for now

package.json
```diff
@@ -11,24 +11,27 @@
     "fix": "prettier --ignore-path .gitignore --write . && eslint . --fix && yarn typecheck",
     "test:run:servers": "sudo pwd && concurrently \"cd ui/ && sudo docker-compose up\" \"npm run test:run --prefix ui\"",
     "test:open": "cypress open",
-    "test:run": "cypress run",
-    "test": "./scripts/wait.sh 3001 && npm run test:run",
+    "test:run": "cypress run --browser chrome --headless",
+    "test": "./scripts/wait.sh 3000 && npm run test:run",
```
Powersource, line 15: why not 3001? the frontend seems to run there

Powersource (commented):
will see if we can get the pipelines greener

Powersource:
@alinauroz have you gotten the tests to run in CI? they seem to run for a bunch of hours and then abort

package.json
Powersource: had to change this back to get it to work in CI

Powersource (changes requested):
I got it to run in CI now but it seems to break when registering a user/logging in

cypress/e2e/bucket/filter-bucket.cy.js
alinauroz: Sure, I'll replace waits

alinauroz:
The issue seems to be with .env.test.

I pulled the changes, copied .env.local to .env.test and ran the test. They are working fine.

aerugo:
> The issue seems to be with .env.test.
> 
> I pulled the changes, copied .env.local to .env.test and ran the test. They are working fine.

@gustavlrsn is something missing in the setup on Github?

Powersource:
Tests will need to be fixed in CI or they won't be very useful to us

Issue #391, opened 2021-11-05
[FEATURE] Substring search for buckets in collection


aerugo:
Buckets are currently not searchable by substring. They should be.

PR #916, opened 2022-12-12
Bucket search

alinauroz:
for https://github.com/cobudget/cobudget/issues/391


alinauroz:
I also added debounce search on change on search input. There is max 1 api call in 300ms. Is that ok? 

@aerugo

gustavlrsn (approved):

Issue #960, opened 2023-02-20
[FEATURE] Submit expenses


aerugo:
When a project has been funded, co-creators of the bucket should be allowed to submit expenses.

_This issue describes the default behaviour, when Open Collective is not connected to the round.
If Open Collective is connected to the round, expense objects are still created on Cobudget but all the data and payment status is read from their API and saved on Open Collective using OAuth of the users own account on Open Collective. A separate issue will describe the implementation when Open Collective integration is enabled. _

Once a bucket reaches status "Funded", a new tab "Expenses" is available on the bucket page.

<img width="746" alt="Screenshot 2023-02-20 at 00 44 08" src="https://user-images.githubusercontent.com/7785081/219984247-e1d8230f-bcdd-4d85-be5e-2e04ae1bfaeb.png">

When a co-creator submits an expense, they can upload one or multiple receipts. Implement drag and drop upload for receipts. 

<img width="910" alt="Screenshot 2023-02-20 at 01 15 17" src="https://user-images.githubusercontent.com/7785081/219985856-361a0202-bc7f-4374-aac2-f98855b6cd77.png">

When opening an expense from the list by clicking on its title, a popup modal showing a table view of the expense is shown.

<img width="763" alt="Screenshot 2023-02-20 at 01 41 06" src="https://user-images.githubusercontent.com/7785081/219986008-b42ca5e1-9714-4061-b5d6-79ffde39feef.png">

Expense owners, admins and moderators see all the information.

Members of the round who do not own the expense do not see the payment details, unless they are admins.

Admins and moderators can change the payment status by clicking the status label and selecting a new status. Similar implementation on Open Collective shown below.

<img width="231" alt="Screenshot 2023-02-20 at 01 48 30" src="https://user-images.githubusercontent.com/7785081/219986212-d11c4ca8-e0a2-4d85-91d7-8ad71aad17d3.png">

alinauroz:
@aerugo 

Adding multiple receipts is given to the user but there is only 1 option to enter payment method. Is this expected behaviour?

What is 1 receipt is paid through SWIFT/IBAN, other through paypal, another through wise?

<img width="816" alt="Screenshot 2023-03-01 at 11 41 03 PM" src="https://user-images.githubusercontent.com/38759611/222234611-f3445296-4c91-416a-a61d-919b0acc019f.png">

aerugo:
[...]

PR #980, opened 2023-04-02
Feat/expenses

alinauroz:
for #960


alinauroz:
1. Co-creators can add and edit expenses/receipts
2. Admin and superamin can view details and change the expense status
3. Other users can just view expenses (not the receipts)

Issue #988, opened 2023-04-11
[MINOR]  "lock" expense data field information


DavidsBees:
**Describe the idea**
Lock the data that is written in the expense field until completed - alternatively prohibit closing the window accidentally before data is logged in.

Data disappears once i click next to the window

![image](https://user-images.githubusercontent.com/80971095/231187575-e6f26b64-7bca-414e-8247-eca913827752.png)

PR #1011, opened 2023-05-11
Remove expense onclose

alinauroz:
fixes https://github.com/cobudget/cobudget/issues/988


aerugo:
@alinauroz - this fails e2e tests, any idea why?

PR #1013, opened 2023-05-14
Don't close expense modal

aerugo:


aerugo:
@alinauroz - this fails

PR #1202, opened 2024-04-16
Feat/join dream

rootux:
# What
- Join dream button
- Email to dreamer should include the commenter email

# Tophatting 🎩 
Tested that it does adds a new comment of Joined dream

Screenshot 2024-04-16 at 21 17 16
Screenshot 2024-04-16 at 21 18 00


rootux:
Oops... wrong repository :)

Issue #1232, opened 2025-10-06
[BUG] Round member limits not enforced correctly upon self joining


rozagerardo:
**Describe the bug**
The Roud limits are not enforced correctly. The Free limit is used correctly for invitations, but users can join the round directly even if the ound is over the limit. Some parts are using the Paid limit instead (for Rounds associated to the "c" Root Group), and some others are not checking the limit at all

**To Reproduce**

1. Create a round
2. Create users and keep joining them the round, there is no limit

**Expected behavior**
When the round reaches its limit, it shouldn't show the "Join Round" button nor allow the functionality if invoked (throw an error).

**Screenshots**

**Browser and device (please complete the following information):**

- Browser [e.g. chrome, safari], and version if applicable
- Device and/or OS: [e.g. iPhone6, Windows, Mac]

**Additional context**

PR #1233, opened 2025-10-06
Enfore correct round member limits upon self joining

rozagerardo:
Issue: https://github.com/cobudget/cobudget/issues/1232


Powersource (approved):
I don't understand all of this anymore but I feel it's safe enough to merge :)

PR #1237, opened 2025-10-26
[GH-1236] Configuring self-set pricing subscription

rozagerardo:
Addresses: https://github.com/cobudget/cobudget/issues/1236

- Taking configured prices from Stripe product instead of from two env variables
- updated upgrade message banner as per specs
- Using select for choosing price in new group, in upgrade group and in upgrade message banner
- disabling bucket and round actions when round has reached limit

Required changes:
- use `STRIPE_PRODUCT_ID` env property
- optionally remove `STRIPE_MONTHLY_PLAN_PRICE_ID` and `STRIPE_YEARLY_PLAN_PRICE_ID`

Screenshot 2025-10-26 010612
Screenshot 2025-10-26 010543
Screenshot 2025-10-26 010417
Screenshot 2025-10-26 010341
Screenshot 2025-10-26 010720
Screenshot 2025-10-26 011111


francesca-sp:
### Update requests 

We are missing the text that can be seen in this wireframe, to inform group and round admins that there is a 10 user limit. It's important they see this from the beginning, so they know what to expect. 

This is supposed to go on the Members list page of groups, and the Participants list page for rounds. 

image

